Issue #26, opened 2015-02-26
rewrite unit tests


bmcfee:
... since the schema's all different, and much of the (python) codebase has been simplified or rewritten, this is probably necessary.

@ejhumphrey Any objection to switching from `unittest.TestCase` objects (old and busted) over to nosetest functions (new hotness)?  This will make tests easier to write, and play nicely with #25 .

bmcfee:
Sketching out a plan for test design:

- [ ] JObject
  - [ ] serialization, deserialization
  - [ ] equivalence testing
  - [ ] dict-like methods
  - [ ] search
- [ ] Sandbox
  - [ ] just test for construction; all else inherits from JObject
- [ ] JamsFrame
  - [ ] construction, conversion
  - [ ] add_observation
  - [ ] interval_values conversion
  - [ ] serialization (sparse and dense)
- [ ] Annotation
  - [ ] construction
  - [ ] append
- [ ] Curator
  - [ ] construction
- [ ] AnnotationMetadata
  - [ ] construction
- [ ] FileMetadata
  - [ ] construction
- [ ] AnnotationArray
  - [ ] construction
  - [ ] search
  - [ ] serialization
- [ ] JAMS
  - [ ] construction
  - [ ] load
  - [ ] save
  - [ ] add
  - [ ] validate
- [ ] pyjams functions
  - [ ] load
  - [ ] append
  - [ ] load_lab
- [ ] Utilities
  - [ ] timedelta_to_float
  - [ ] serialize_obj (should be implicit from JamsFrame tests)
  - [ ] read_lab (do we still need this?)
  - [ ] load_textlist
  - [ ] expand_filepaths
  - [ ] smkdirs
  - [ ] filebase
  - [ ] find_with_extension
  - [ ] match_query (semi-redundant with JObject.search test)
  - [ ] query_pop (also semi-redundant with JObject.search test)

urinieto:
Nosetests +1

This test list is.... well, I have no words, I'm so impressed.

bmcfee:
> This test list is.... well, I have no words, I'm so impressed.

save being impressed for after we've implemented it!

ejhumphrey:
sounds great; any preferred examples to share of this in practice? or any
ole google search will do?

bmcfee:
> any preferred examples to share of this in practice?

nose is super easy to use, but librosa has tons of (not so clean) [examples](https://github.com/bmcfee/librosa/blob/develop/tests/test_met_features.py)

[...]

Issue #16, opened 2015-01-31
mir_eval integration


bmcfee:
Quick thoughts about making integration with mir_eval simple:

* Provide functionality to generate and populate Annotation objects from .lab files
* Provide wrapper functions for all mir_eval submodules
  * This can be implemented via a handful of transparent decorators
  * The decorated evaluators accept two Annotation objects `(ref_annotation, est_annotation)`, does the necessary unpacking into `mir_eval` array/list format, and returns the result
  * We may need multiple decorators: events, labeled events, intervals, and labeled intervals.  Possibly more may be necessary depending on the task/metric.
  * We may also need to do docstring mangling to change the function signature while retaining the description.

bmcfee:
Other functions we'll need

* `lab_to_jam(namespace, filename, *parsing)`
  * create a new jam object by importing a lab file

bmcfee:
> lab_to_jam(namespace, filename, *parsing)

first cut as of e5e13321577a33ac26adea1969d3a7d9cf5df7ee

bmcfee:
776c0876b67821e613a216010e16bad95a95da69 implements a first cut at the decorator-wrapping for mir_eval functions.

The gist of this is that for each mir_eval submodule (beat, chord, onset, etc), we wrap the master "evaluate" function which calls all evaluation metrics and returns an ordered dict of scores.  Since different tasks have different conventions for data (onsets just have event time, chords have interval labels, etc), we need a few different wrappers to make everything work.

TODO:
  * Fix the call signatures and docstrings
  * Add source separation.  This still needs a #13 namespace schema before we can implement the evaluator.
  * Add pattern discovery.  Again, we need a namespace.

We can have the decorator verify that the input annotations have the right type for the evaluation in question.  
  * For something like segment, the estimation might disagree with the reference (`segment_open` vs `segment_tut`, for instance), but that's ok if we search on 'segment_*'.
  * Will this pattern hold in general?  Can we just make all evaluation wrappers validate on the pattern '^TASK(_.*)?$' ?

bmcfee:
Finally started implementing tests for the eval module #26 ; this looks to be pretty badly broken.

The event-oriented evaluation wrappers (beat and onset) work, since their call signatures fit with `mir_eval`.  The others (chord, segment, tempo) do not work, since the wrapped versions take two annotations as input, while mir_eval takes more than two input arguments (eg, `ref_intervals, ref_labels, est_intervals, est_labels`).

[...]

PR #42, opened 2015-06-23
Pattern eval

bmcfee:
This PR implements mir_eval integration for pattern discovery.  Ref #16 and #26


bmcfee:
@urinieto I think this one's ready to go.  Care to give it a once-over before merging?

Issue #66, opened 2015-09-15
Partial annotations


craffel:
It would be great if JAMS could seamlessly support "partial annotations", by which I mean annotations where the annotator has only annotated a portion of a song.  For example, say someone annotates chords in a song from 10s to 30s.  It would be possible to recover the annotated range by looking at `ann.data.time.min()` and `(ann.data.time + ann.data.duration).max()`, which is a little clunky but would work.  However, if, say, onsets (or other 0-duration events) were annotated instead, this method would effectively treat the annotated range as `ann.data.time.min()` to `ann.data.time.max()`.  If evaluation was then performed over this range, it would ignore the ground-truth judgement made by the annotator that there are no onsets between `ann.data.time.max()` and 30s (in this example).  So, it'd be nice to have some way to codify that an annotator has annotated a specific range.

This issue came up recently when thinking about using aligned MIDI files for ground truth.  It's common that alignment will only be successful for a subset of a song (perhaps because of a poor transcription of part of the song, or structural differences, etc), so in most cases aligned MIDI files produce "partial alignments".  Onsets and beats are both 0-duration annotations which are trivially easy to extract from MIDI files, and it would be great to evaluate over the entire "correctly aligned" range, rather than the range of the annotations.

bmcfee:
I think this is a great idea.

A simple fix which requires no effort on our part is to stash this information in the annotation's sandbox.  This is what the sandbox is for: storing things that the jams devs didn't think of. :)  It's possible to query/search on sandboxes, so if you wanted to find annotations that hit a particular time range, it should be fairly easy to do.

However, this seems like a good use case that should be codified and brought into the jams schema.  I'm thinking that adding `time` and `duration` fields to the `Annotation` object would do the trick.  For backwards compatibility, these can be optional, and if left blank, assumed to cover the entire track.

Anyone else care to chime in?

justinsalamon:
:+1:

craffel:
>  I'm thinking that adding time and duration fields to the Annotation object would do the trick. For backwards compatibility, these can be optional, and if left blank, assumed to cover the entire track.

This seems like a good solution.

bmcfee:
To-do:

- [ ] Add `time` and `duration` fields to the Annotation object schema.
- [ ] Corresponding tests, documentation, and examples.

bmcfee:
Implemented by merging #67

PR #67, opened 2015-09-15
Partial annotation [implements #66 ]

bmcfee:
Implements #66

### DIFF
--- a/docs/examples/example_beat.py
+++ b/docs/examples/example_beat.py
@@ -36,7 +36,7 @@ def beat_track(infile, outfile):
     jam.annotations.append(beat_a)
 
     # Add tempo estimation to the annotation.
-    tempo_a = jams.Annotation(namespace='tempo')
+    tempo_a = jams.Annotation(namespace='tempo', time=0, duration=track_duration)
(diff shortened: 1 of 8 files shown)


bmcfee:
I've implemented the schema changes as described in #66 .  A couple of things to hammer out before merging:

- Do we want the time/duration fields to be timedeltas (as in the jamsframe) or plain floats?  The latter is simpler and fits with the FileMetadata duration field, but timedeltas would make it fit with jamsframe.

- Should we make them required fields?  Right now it's either a non-negative number or 'null'.  Is this okay with everyone?

bmcfee:
Since nobody's taking my bait, here's my 2e-2€:

> Do we want the time/duration fields to be timedeltas (as in the jamsframe) or plain floats? The latter is simpler and fits with the FileMetadata duration field, but timedeltas would make it fit with jamsframe.

I vote `float` for simplicity and consistency with file_metadata.  (This is a pyjams thing, not a schema thing.)

> Should we make them required fields? Right now it's either a non-negative number or 'null'. Is this okay with everyone?

I don't think we should make them required.


If everyone agrees with me on the above, this PR is ready to merge.

justinsalamon:
:+1:

urinieto:
Sounds good to me too.

Issue #70, opened 2015-09-20
Search in sandbox


urinieto:
I am trying to search an annotation that has a specific value in its sandbox, but I'm not sure how to use the `search` function.

If the sandbox field looks like this:

```"sandbox": {
    "test": "titan"
}```

How should I call the `search` function in order to get this annotation?

Thanks!

**Comment by bmcfee** [comment]
https://gist.github.com/2513087f4f350f52e43a shows an example of how to do this.

The complicating factor here is that searching on nested objects requires some strange syntax, ie, the `.` operator.  This screws up python's keyword-argument parsing if you do it directly, say `jam.search(Sandbox.foo='bar')`, but if you use the ** magic, everything works as expected.

**Comment by bmcfee** [comment]
Follow-up note: you could also search on the bare field name directly:
```python
annotatoins = jam.search(test='titan')
```
and that would also work.  If you have unique field names that don't collide with anything else in the `Annotation` object hierarchy (eg, `AnnotationMetadata`,  `Curator`, or `JamsFrame`), this might be preferable.

urinieto:
Awesome, it works. It's confusing, since "sandbox" in the JAMS file itself does not have an upper case 'S', but I guess it's looking for the actual object, which is capitalized.

Regardless, thanks titan!

urinieto:
Mmhh I have another question: how do I make joint queries with this syntax? Let's say I want to find the same sandbox test _and_ `namespace="segment_salami_upper"`. What would be the best way to do it?

Thanks again!

bmcfee:
> Awesome, it works. It's confusing, since "sandbox" in the JAMS file itself does not have an upper case 'S', but I guess it's looking for the actual object, which is capitalized.

Right, because it's searching over the type, not the instance variable itself.  This may seem weird, but it's the only way to be consistent when you can have multiple objects of the same type, as is the case with an array of annotations.  Using the class name anonymizes the object.


> Mmhh I have another question: how do I make joint queries with this syntax? Let's say I want to find the same sandbox test and namespace="segment_salami_upper". What would be the best way to do it?

By 'joint' you mean 'conjunctive'.  JAMS queries are disjunctive by default, so you can do OR queries directly:
```python
results = jam.search(**{'namespace': 'foo', 'Sandbox.test': 'titan'})
```

If you want a conjunction (AND) query, you can post-process the results array.  It's an AnnotationArray type, and not a vanilla list, so the following would work:
```python
[...]

PR #71, opened 2015-09-21
implemented object equality search

bmcfee:
Simplifies the search interface for non-string targets.  Created in response to the comments in #70 .


urinieto:
Woah, nice!

Issue #55, opened 2015-08-20
local namespace repository


bmcfee:
Dynamically adding custom namespaces on import is a real drag.

Instead, we should allow custom namespaces to be stored in a directory specified by an environment variable or rc file.

bmcfee:
Implemented in #75

PR #75, opened 2015-10-18
Local namespaces

bmcfee:
This PR implements #55 by allowing the user to specify a path to custom JAMS namespace schema via the environment variable `JAMS_SCHEMA_DIR`.


bmcfee:
For simplicity, I skipped implementing the config file option.  Also, only a single custom path directory is supported.

Does anyone foresee a need for supporting any of the following features?

- system-wide config file
- user config file
- multiple custom schema directories

If not, I think this is merge-ready.  CR kindly requested, but I think it's pretty straightforward.

PR #94, opened 2016-01-28
fixed a type error in jams_to_lab

bmcfee:
jams_to_lab didn't actually work because ``Annotation`` objects ceased being hashable.

This PR fixes the conversion logic to use an array rather than a set.

Anyone care to CR?  It should be an easy one.  (Maybe @ejhumphrey )

### DIFF
--- a/scripts/jams_to_lab.py
+++ b/scripts/jams_to_lab.py
@@ -124,10 +124,10 @@ def convert_jams(jams_file, output_prefix, csv=False, comment_char='#', namespac
     # Make a counter object for each namespace type
     counter = collections.Counter()
 
-    annotations = set()
+    annotations = []
 
     for query in namespaces:
-        annotations.update(jam.search(namespace=query))
+        annotations.extend(jam.search(namespace=query))
 
     if csv:
         suffix = 'csv'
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/marl/jams/pull/94" target="_blank">ReviewNinja

scripts/jams_to_lab.py
```diff
@@ -124,10 +124,10 @@ def convert_jams(jams_file, output_prefix, csv=False, comment_char='#', namespac
     # Make a counter object for each namespace type
     counter = collections.Counter()
 
-    annotations = set()
+    annotations = []
 
     for query in namespaces:
```
ejhumphrey: Am I crazy or is this going to fail with the default args (`namespaces=None`)?

bmcfee: Good call.  I'll throw an exception if namespaces is None.

scripts/jams_to_lab.py
```diff
@@ -124,10 +124,10 @@ def convert_jams(jams_file, output_prefix, csv=False, comment_char='#', namespac
     # Make a counter object for each namespace type
     counter = collections.Counter()
 
-    annotations = set()
```
ejhumphrey: If we're worried about uniqueness, it'd probably be safer to just dump all the annotations with `namespace` suffixes -- afaik, this is fine

scripts/jams_to_lab.py
bmcfee: !resolved

scripts/jams_to_lab.py
bmcfee: !resolved

PR #96, opened 2016-01-31
RFC First pass at jamongo API

cjacoby:


jams/jamongo.py
```diff
@@ -0,0 +1,171 @@
...
+
+This library provides an interface for interfacing with a JAMS Mongo
+instance.
+
+"""
+
+import pymongo
+
+
+def convert_annotation_list(annotations_array):
```
bmcfee: 8-space tabs! Copypasta

jams/jamongo.py
```diff
@@ -0,0 +1,171 @@
...
+        return self
+
+    def __exit__(self, type, value, traceback):
+        pass
+
+    @property
+    def db(self):
+        return self._db
+
+    @property
```
bmcfee: We have "annotations" and "audio" here... should there be somewhere that explicitly defines these collections?  And should these getters be attributes, or would it make more sense to provide a dict interface in this object where the key maps to a collection in `_db`?

jams/jamongo.py
```diff
@@ -0,0 +1,171 @@
...
+        jams_object : jams.JAMS
+        """
+        # Extract the JAMS metadata & insert it into the Audio collection
+        audio_id = self.insert_jams_metadata(jams_object.file_metadata)
+
+        # Extract the JAMS Annotations and prepare them for mongification.
+        annotation_list = convert_annotation_list(jams_object.annotations)
+
+        # insert them into the Annotations collection.
+        annotation_ids = self.insert_annotations(annotation_list, audio_id)
```
bmcfee: Maybe simpler to have a method that inserts one at a time, and then iterate over the annotation_list?

jams/jamongo.py
cjacoby: drp. fixed it.

jams/jamongo.py
cjacoby: re: "should there be somewhere that explicitly defines these collections":
+1 on that.

re: "And should these getters be attributes, or would it make more sense to provide a dict interface in this object where the key maps to a collection in _db"

Do you mean something like:
```self._collections = ["music", "annotations"]

def __getattr__(self, key):
    return self._collections[key]
```

???

jams/jamongo.py
```diff
@@ -0,0 +1,183 @@
...
+    def db(self):
+        return self._db
+
+    def __getattr__(self, key):
+        """Get a collection object from the db, if it is an allowable
+        collection specified."""
+        if key in self._collections:
+            return self._db['key']
+        else:
+            return None
```
bmcfee: do we want a `return None` here, or throw a key error?

jams/jamongo.py
bmcfee: !resolved

jams/jamongo.py
cjacoby: ... yeah key error is better.

jams/jamongo.py
bmcfee: !resolved

jams/jamongo.py
bmcfee: !resolved

jams/jamongo.py
```diff
@@ -0,0 +1,183 @@
...
+    Parameters
+    ----------
+    annotations_list : jams.AnnotationArray
+
+    audio_id : ObjectId
+        Pointer to an audio document.
+
+    Returns
+    -------
+    converted_annotations : list of Annotation
```
bmcfee: Just to be 100% clear, we for sure want deep-copied `jams.Annotation` objects returned here, and not simplified dicty json objects.  Right?

cjacoby: Yes. We'll just have to do a lil' old `[x.__json__ for x in annotations]` before we actually insert it, but I think this is what we want.

bmcfee: !resolved

jams/jamongo.py
bmcfee: !resolved

PR #97, opened 2016-01-31
Ejh 20160131 schema mods

ejhumphrey:


jams/schema.py
```diff
@@ -67,6 +67,39 @@ def namespace(ns_key):
     return sch
 
 
+def definition_schema(def_name, raise_if_missing=True):
```
bmcfee: It's probably worth doing this once up front to populate all the object types on load, rather than doing it every time the `__schema__ is called.

ejhumphrey: is this a `plzfix` or a `todo`? :o)

PR #98, opened 2016-01-31
20160131 mango queries

cjacoby:
Added some query functions and unit tests.


bmcfee:
This PR seems dead.  Can we close it?

Issue #28, opened 2015-05-19
Namespace converters


bmcfee:
Wherever it makes sense, we should have converter scripts to move between namespaces.

This could get pretty messy.

From the current set, the following seem feasible:

- [ ] `beat_position` -> `beat`
- [ ] `tag_*` -> `tag_open`
- [ ] `segment_*` -> `segment_label_open`
- [ ] `pitch_hz` <-> `pitch_midi`
- [ ] `pitch_class` -> `pitch_midi`, `pitch_hz`
- [ ] `chord_roman` -> `chord_harte` (?)

(I'm not sure about that last one; is it possible?)

There are some issues to hammer out with regards to additional parameters of the conversion.  For instance, `pitch_class_to_midi` would probably require an octave indicator, unless we just make everything start at C4 or something.

Of course, all of this is up for discussion.

bmcfee:
Some ideas:

* Add a method `Annotation.to_namespace(target_namespace)`
* Add custom converter functions as submodules of `jams.ns`.
* Register converter functions with a dispatch table keyed by tuples `(ns_from, ns_to)`, where `ns_from` is a regexp to match the input annotation namespace, and `ns_to` is the `target_namespace`

When the dust settles, an example converter might look like:

```python
@converter('tag_.*', 'tag_open')
def to_tag_open(annotation):
    annotation.namespace = 'tag_open'
    return tag_open
```

The `converter` decorator handles the following:
* registering the decorated function with the source and target namespaces in the dispatch table
* validating input annotation against the source namespace
* copying the annotation object to ensure safety of the original object

bmcfee:
Another idea for #16:

* automatic conversion attempts whenever an input annotation fails the default type check

For instance `jams.eval.chord()` only operates on `chord_harte` annotations.  Currently, if an annotation is not of the right namespace, it fails with an exception.  However, we can instead try:
```python
R = ref_annotation.to_namespace('chord_harte')
E = est_annotation.to_namespace('chord_harte')
```
and proceed as normal.

justinsalamon:
@bmcfee do we have any existing converters I could have a look at? Looking to implement the ``pitch_hz``<->``pitch_midi`` converter asap.

bmcfee:
Nope, I haven't gotten around to it yet, mainly because I haven't had a need yet.

However, having it in place might simplify some of the sonification logic, and I think the decorator syntax makes sense.

[...]

PR #105, opened 2016-02-16
[WIP] namespace conversion

bmcfee:
Implements #28 

- [x] `beat_position` -> `beat`
- [x] `tag_*` -> `tag_open`
- [x] `segment_*` -> `segment_label_open`
- [x] `pitch_hz` <-> `pitch_midi`
- [x] `chord_harte` -> `chord`
- [ ] ~~~`pitch_class` -> `pitch_midi`, `pitch_hz`~~~
- [ ] ~~~`chord_roman` -> `chord_harte` (?)~~~

### DIFF
--- a/jams/eval.py
+++ b/jams/eval.py
@@ -28,7 +28,7 @@ from .nsconvert import convert
 __all__ = ['beat', 'chord', 'melody', 'onset', 'segment', 'hierarchy', 'tempo', 'pattern']
 
 
-def validate_annotation(ann, namespace):
+def coerce_annotation(ann, namespace):
     '''Validate that the annotation has the correct namespace,
     and is well-formed.
 
@@ -101,8 +101,8 @@ def beat(ref, est, **kwargs):
     '''
 
     namespace = 'beat'
-    ref = validate_annotation(ref, namespace)
-    est = validate_annotation(est, namespace)
+    ref = coerce_annotation(ref, namespace)
+    est = coerce_annotation(est, namespace)
     ref_interval, _ = ref.data.to_interval_values()
     est_interval, _ = est.data.to_interval_values()
 
@@ -142,8 +142,8 @@ def onset(ref, est, **kwargs):
     >>> scores = jams.eval.onset(ref_ann, est_ann)
     '''
     namespace = 'onset'
-    ref = validate_annotation(ref, namespace)
-    est = validate_annotation(est, namespace)
+    ref = coerce_annotation(ref, namespace)
+    est = coerce_annotation(est, namespace)
     ref_interval, _ = ref.data.to_interval_values()
     est_interval, _ = est.data.to_interval_values()
 
@@ -184,8 +184,8 @@ def chord(ref, est, **kwargs):
     '''
 
     namespace = 'chord'
-    ref = validate_annotation(ref, namespace)
-    est = validate_annotation(est, namespace)
+    ref = coerce_annotation(ref, namespace)
+    est = coerce_annotation(est, namespace)
     ref_interval, ref_value = ref.data.to_interval_values()
     est_interval, est_value = est.data.to_interval_values()
 
@@ -226,8 +226,8 @@ def segment(ref, est, **kwargs):
     >>> scores = jams.eval.segment(ref_ann, est_ann)
     '''
     namespace = 'segment_open'
-    ref = validate_annotation(ref, namespace)
-    est = validate_annotation(est, namespace)
+    ref = coerce_annotation(ref, namespace)
+    est = coerce_annotation(est, namespace)
     ref_interval, ref_value = ref.data.to_interval_values()
     est_interval, est_value = est.data.to_interval_values()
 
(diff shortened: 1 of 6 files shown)


bmcfee:
@justinsalamon give it a look.  I'll whip up some tests shortly, but all you have to do now is say
```python
>>> jams.convert(annotation, 'pitch_hz')
```
if such a conversion is possible, it will be done.  If not, you get an exception.

It's designed to make adding new converters pretty easy (see the `nsconvert.py` module).

justinsalamon:
@bmcfee just had a look (hence ninja star), looks good to me.

bmcfee:
Okay, cool.  I'll get on with testing then.

I'll also rework eval a bit to use the converters where appropriate.

justinsalamon:
Sounds good, ping me if you'd like me to have a second look, and hopefully we can merge this pretty soon.

bmcfee:
Ok, I think this one's ready to go.

Give it a last CR pass and I'll merge.

jams/eval.py
```diff
@@ -54,10 +54,7 @@ def validate_annotation(ann, namespace):
         If `ann` fails schema validation
     '''
 
-    if not ann.search(namespace=namespace):
-        raise NamespaceError('Expected namespace="{:s}", found "{:s}"'
-                             .format(namespace, ann.namespace))
-
+    ann = convert(ann, namespace)
```
justinsalamon: @bmcfee am I right in understanding that you're validating the jams by converting it to the desired namespace and then checking if it validates against this namespace?

If so, doesn't this mean you're not only doing validation, but also conversion, under the hood?

bmcfee: The test was previously

"Is this annotation of namespace `X`?"

and is now

"can this annotation be interpreted as a valid instance of namespace `X`".


I think for evaluation (and sonification) the latter is what we really want.

bmcfee:
> for example here, does this mean you would validate a chord_harte annotation by trying to convert it to the chord namespace and check if it validates there? Is this safe?

This is safe because `chord_harte` is a (proper) subset of `chord`.  Going the other direction would not be safe in general.  (Similarly for conversion between closed tag or segment vocabularies.)

It wouldn't be hard to do a best-effort reverse conversion (`chord` -> `chord_harte`) by clobbering the namespace and calling validate.  If it passes, we're good.  If it fails, reraise the exception saying that you have an uncastable input (eg, a `sus` chord).

I'd prefer to not go down that rabbit hole though, since it opens up a lot of possibilities for *attempted* conversion.  Better to stick with a set of well-defined converters that can keep guarantees about the output.

jams/eval.py
justinsalamon: Right. So let's say I have a ``pitch_midi`` annotation, and I validate it against the ``pitch_hz`` namespace, which would validate fine in this new scenario. So I then try to use this annotation with the transcription eval wrapper, which expects ``pitch_hz`` annotations, and it breaks.

Now of course in practice I'll be adding a conversion step to the transcription eval wrapper such that any annotation provided is converted into ``pitch_hz``, so this wouldn't actually happen. But could this type of scenario, where you validate against a namespace different to the one your annotation actually is (but convertible to) and consequently get errors further down the line somewhere?

bmcfee: > So I then try to use this annotation with the transcription eval wrapper, which expects pitch_hz annotations, and it breaks.

Ah, I see what you're saying.  I forgot to change the return type.

Perhaps this should be renamed to `coerce_annotation`.

jams/eval.py
```diff
@@ -52,15 +55,16 @@ def validate_annotation(ann, namespace):
...
+    See Also
+    --------
+    jams.nsconvert.convert
+    '''
 
+    ann = convert(ann, namespace)
     ann.validate(strict=True)
 
-    return True
+    return ann
```
justinsalamon: This look safer, yes

jams/eval.py
bmcfee: Ok, I've updated the validation logic.

Instead of returning `true` (which was never actually checked) it now returns the coerced annotation, which is what gets piped into mir_eval.

jams/eval.py
```diff
@@ -32,6 +32,9 @@ def validate_annotation(ann, namespace):
     '''Validate that the annotation has the correct namespace,
```
justinsalamon: Should we be concerned about potential confusion between `eval.validate_annotation()` and `annotation.validate()` ?

jams/eval.py
justinsalamon: Yes I just had a look. Already added a couple of line comments for your attention.

jams/eval.py
bmcfee: It's not (really) a public-facing function, so I don't think so.

But, like I said, `coerce_and_validate` might be a more accurate description.

jams/eval.py
bmcfee: !resolved

jams/eval.py
bmcfee: Actually, it's not included in `__all__` or the docs, so it's 100% private and we can do whatever we want without worrying about breaking API.

Shall I change it?

justinsalamon: Well, just so that the function name reflects the functionality, I'd say yes (`coerce_and_validate`)

bmcfee: Done.

justinsalamon: Cool. You might want to update the docstring of the function, but otherwise everything looks good.

PR #119, opened 2016-07-04
added namespace enumerated values accessor

bmcfee:
This PR adds an accessor function for enumerated namespaces (eg, `tag_gtzan`).

Example usage:

```python
>>> values = jams.schema.values('tag_gtzan')
['blues', 'classical', 'country', ...]
```

This is useful for supervised learning tasks, where it's helpful to know the vocabulary up front (eg, for building label encoders).

It only works/makes sense if the namespace's `values` field is an `enum` type.  This rules out things like `tag_open` or `chord`, which are defined by regular expressions (and are, in effect, infinite vocabularies).  If the given namespace is not enumerated, a `NamespaceError` is thrown.  Otherwise, a copy of the values array is returned.

This should be ready for CR.

### DIFF
--- a/jams/schema.py
+++ b/jams/schema.py
@@ -10,6 +10,7 @@ Namespace management
     add_namespace
     namespace
     is_dense
+    values
 '''
 
 import json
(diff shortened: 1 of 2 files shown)


bmcfee:
@ejhumphrey CR plz?

bmcfee:
crickets

jams/schema.py
```diff
@@ -88,6 +88,33 @@ def is_dense(ns_key):
     return __NAMESPACE__[ns_key]['dense']
 
 
+def values(ns_key):
+    '''Return the allowed values for an enumerated namespace.
```
rabitt: An example here would be helpful.

rabitt:
👍 Everything looks good. Only thing I changed was I added the gtzan example to the values function docstring.

bmcfee:
> Only thing I changed was I added the gtzan example to the values function docstring.

thanks! I updated the doc build to include this and fixed a typo; otherwise this is mergeable.

Issue #112, opened 2016-05-06
Dictionary interface for quick access to annotations


bmcfee:
I was chatting with Peter today about some of the more annoying aspects of using jams for simple tasks.  For example, to get a single `onsets` annotation out of the jams object, one has to do something like the following:

``` python
>>> anns = jam.search(namespace='onsets')
>>> onset = anns[0]
...
```

which is rather cumbersome.

I think we can shortcut this a little by providing a read-only dictionary interface to the jams object:

``` python
>>> onset = jam['onset'][0]
```

where the `JAMS` object under the hood just short-circuits the search for us:

``` python
def __getitem__(self, query):
    return self.search(namespace=query)
```

Would this level of syntactic sugar be helpful?

justinsalamon:
Yes

ejhumphrey:
wholly agree! 

I'd challenge us to take it one step further. I find it really unnerving that annotations are order sensitive. If I interact with a jam the right way (intentionally or otherwise), I can end up with annotations taking on new index values (e.g. shuffling, slicing, etc). 

What if we adopted the pandas conventions of `.loc` and `.iloc` (or different names, I don't feel strongly), so that you can specify an index for annotations in the whole of a jam (I realize this might involve a schema change, but hang with me for a second). For example:

```
>>> ann1 = jams.Annotation('beat', data=[dict(time=0.2)])
>>> ann2 = jams.Annotation('onset', data=[dict(time=2.5)])
>>> jam = JAMS([ann1], index=['foo'])
>>> jam.append(ann2, index='bar')
>>> jam.loc['foo'].data
             time  duration  value  confidence
0 00:00:00.200000       NaT    NaN         NaN

>>> jam.iloc['onsets', 0].data
             time  duration  value  confidence
0 00:00:02.500000       NaT    NaN         NaN

```

not a fully formed idea, but curious to field thoughts.

bmcfee:
> I find it really unnerving that annotations are order sensitive.

Agreed.  I blame javascript for not having a set primitive data structure.

The obvious way around this is to provide each annotation with a key / index, like you describe.  This could be done by replacing `AnnotationArray` with an `AnnotationDict` (and the corresponding schema change).  The downside to this is the `index` would be part of the annotation dict object, and not the annotation itself; this means it will be harder to separate annotations from jams as described in #92 .

[...]

PR #120, opened 2016-07-09
added fancy annotationarray indexing

bmcfee:
Implements #112

- [x] documentation
- [x] unit tests

### DIFF
--- a/jams/core.py
+++ b/jams/core.py
@@ -995,7 +995,10 @@ class AnnotationArray(list):
 
     This list subclass provides serialization and search/filtering
     for annotation collections.
+
     """
+    # TODO
+    #   document fancy indexing
     def __init__(self, annotations=None):
         """Create an AnnotationArray.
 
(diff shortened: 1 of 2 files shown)


bmcfee:
I think this is ready for CR.

@ejhumphrey ?

bmcfee:
tumbleweed

jams/core.py
```diff
@@ -1040,6 +1057,18 @@ def search(self, **kwargs):
 
         return results
 
+    def __getitem__(self, idx):
```
rabitt: tricksy ;)

jams/core.py
```diff
@@ -995,6 +995,23 @@ class AnnotationArray(list):
 
     This list subclass provides serialization and search/filtering
     for annotation collections.
+
+    Fancy-indexing can be used to directly search for annotations
+    belonging to a particular namespace.
+
+    Examples
```
rabitt: These examples are helpful - is there a reasonable place where we could document more generically what kind of indexing is allowed?

rabitt:
This lgtm. My only comment would be to add documentation somewhere that outlines exactly what kinds of indexing can be done, both generically and with the examples you already give.

bmcfee:
Thanks for the CR.  I expanded the docs a little, should be good to go.

Issue #111, opened 2016-04-26
Split pitch_ namespaces into note_ and contour_


justinsalamon:
As discussed in #91 starting [here](https://github.com/marl/jams/pull/91#issuecomment-214840586), currently the `pitch_hz` and `pitch_midi` namespaces are overloaded in a bad way, supporting both note-style annotations (where a note has a start time, duration, and pitch) and continuous f0 (contour)-style annotations (where contours are specified by timestamped pitch values with 0 duration).

Apart from being confusing, this causes a number of problems e.g. for sonification and visualization.

The proposed solution is to split "pitch" into "note" and "contour", keeping the hz/midi distinction as well:
1. `note_hz`
2. `note_midi`
3. `contour_hz`
4. `contour_midi`
Note that the only difference between hz and midi here is how pitch values are represented (on a Hertz scale or as MIDI note numbers), but the two are otherwise equivalent: neither implies any sort of pitch quantization, and both can be used to represent sparse (note) annotations and dense (contour) annotations.

Finally, we need to decide whether to use the aforementioned namespace names as is or add a `pitch_`  prefix to each of the 4 (the former means shorter/cleaner namespace names, the latter is less ambiguous).

bmcfee:
I think i'd err on the side of shorter namespaces here; there isn't a whole lot of ambiguity IMO.


[Tagging @rabitt here.]

Should we try to distinguish between monophonic and polyphonic annotations?  This isn't a big deal for `note_*` which are, for all intents and purposes, piano roll encodings.  It does seem to be a problem for `contour_*` though, when it comes to interpreting the annotations (eg through viz or sonify).

Do we need a separate set of namespaces for multi-f0?  Or just use a collection of annotations to code multi-f0 annotations?  The latter seems a little clumsy to me, but I don't work with this kind of data so it's hard to tell.

A new namespace for multi seems sensible, but I don't know how to structure it. Maybe each observation is a tuple of `(contour_id, frequency_value)`?  Also seems clumsy.

Packing multiple contours into a single `contour_hz` annotation seems like a total nightmare though.  Am I wrong?

justinsalamon:
> I think i'd err on the side of shorter namespaces here; there isn't a whole lot of ambiguity IMO.

Ok, so lets go with:
1. note_hz
2. note_midi
3. contour_hz
4. contour_midi
(the contour ones might change depending on how we decide to handle multi-f0)

[...]

PR #121, opened 2016-07-11
Pitch refactor

bmcfee:
implements #111

### DIFF
--- a/jams/schemata/namespaces/pitch/note_hz.json
+++ b/jams/schemata/namespaces/pitch/note_hz.json
@@ -1,0 +1,8 @@
+{"note_hz":
+    {
+        "value": { "type": "number" },
+        "dense": false,
+        "description": "Note pitches in Hz"
+    }
+}
+
--- a/jams/schemata/namespaces/pitch/note_hz.json
+++ b/jams/schemata/namespaces/pitch/note_hz.json
@@ -1,6 +1,6 @@
 {"note_hz":
     {
-        "value": { "type": "number" },
+        "value": { "type": "number", "minimum": 0},
         "dense": false,
         "description": "Note pitches in Hz"
     }
--- a/jams/schemata/namespaces/pitch/note_midi.json
+++ b/jams/schemata/namespaces/pitch/note_midi.json
@@ -1,0 +1,8 @@
+{"note_midi":
+    {
+        "value": { "type": "number" },
+        "dense": false,
+        "description": "Note pitches in (fractional) MIDI note numbers"
+    }
+}
+
--- a/jams/tests/namespace_tests.py
+++ b/jams/tests/namespace_tests.py
@@ -202,6 +202,35 @@ def test_ns_tempo_invalid():
     for confidence in [-1, -0.5, 2.0, 'a']:
         yield __test, 120.0, confidence
 
+def test_ns_note_hz_valid():
+
+    ann = Annotation(namespace='note_hz')
+
+    seq_len = 21
+    times = np.arange(seq_len)
+    durations = np.zeros(seq_len)
+    values = np.linspace(0, 22050, seq_len) # includes 0 (odd symmetric)
+    confidences = np.linspace(0, 1., seq_len)
+    confidences[seq_len//2] = None # throw in a None confidence value
+
+    for (t, d, v, c) in zip(times, durations, values, confidences):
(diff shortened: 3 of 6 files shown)


jams/tests/namespace_tests.py
```diff
@@ -264,6 +323,43 @@ def __test(value):
...
+    for (t, d, v, c, i, b) in zip(times, durations, values, confidences, ids, voicing):
+        ann.append(time=t, duration=d, value={'pitch': v, 'id': i, 'voiced': b}, confidence=c)
+
+    ann.validate()
+
+def test_ns_contour_invalid():
+
+    ann = Annotation(namespace='pitch_contour')
+
+    seq_len = 21  # should be odd
```
rabitt: Just curious - why "should be odd"?

jams/schemata/namespaces/pitch/note_hz.json
```diff
@@ -0,0 +1,8 @@
+{"note_hz":
+    {
+        "value": { "type": "number" },
```
rabitt: Looks like this should be
```"value": { "type": "number", "minimum": 0},```

because the following test is failing:
```namespace_tests.test_ns_note_hz_invalid(-23,) ... FAIL```

File: jams/schemata/namespaces/pitch/note_midi.json [hunk 3]
```diff
@@ -0,0 +1,8 @@
+{"note_midi":
+    {
+        "value": { "type": "number" },
```
rabitt: Should we also force midi values to be at least 0?

rabitt:
@bmcfee this lgtm now. Only outstanding question is about whether midi values should be restricted to a range.

jams/schemata/namespaces/pitch/note_midi.json
bmcfee: I suppose if we want the spec to conform to MIDI standards, it should be a 7-bit integer [0-127].  But if we want to support fractional numbers, [0, 128), that breaks things a little.

Maybe @craffel can chime in here?

craffel: If you're allowing fractional notes doesn't that already break things a little?

bmcfee: Yes -- it's more of a question of how far we want to (pitch) bend the rules of MIDI.  Are there accepted practices for that kind of thing?

craffel: Not that I know of.  Is MIDI-(this JAMS format)-MIDI intended to be effectively lossless?  Note that standard MIDI pitch bend range is [-2, +2) semitones, so you'd actually need to do [0, 129) (or [-2, 129)?) to cover all possible pitch bends of MIDI notes.

bmcfee: So that immediately breaks the 0-127 range, which makes me think we should leave the spec as is, and not worry about bounds at all.  (Logarithms are unbounded anyway, yolo.)

rabitt: wfm

bmcfee:
@rabitt @justinsalamon the one thing here that differs from #111 discussion: I didn't implement contours with midi number observations, only hz.  Do we actually need/want that?  It seems rather unnatural for contour annotations.

Otherwise, I think this is ready to go.

justinsalamon:
> I didn't implement contours with midi number observations, only hz. Do we actually need/want that? It seems rather unnatural for contour annotations.

Any reason why Hz feels more "natural" than MIDI? Both are just scaled for representing frequency. If we support both elsewhere, for consistency perhaps we should support both here too?

bmcfee:
1. Hz is an SI unit
2. It feels unnatural to use midi for instantaneous data
3. I think it's better to encourage a standard storage format with converters than many standard data formats.

rabitt:
fwiw, there *are* a few datasets with midi contour annotations (Bach 10 for example)

bmcfee:
But they could be converted, right?

rabitt:
Yep they could. I don't feel strongly either way. Happy to go with whatever y'all decide is best.

rabitt:
@bmcfee Can we merge this?

bmcfee:
I think so, yes.  Unless @justinsalamon strongly objects to not having a midi contour namespace.

PR #125, opened 2016-07-22
added infer_duration flag to import_lab

### LINKED ISSUE #107: Import LAB files from Sonic Visualiser
Hey folks,

when exporting lab-files from Sonic Visualiser (use *.txt for this instead of *.csv),
they export in this structure:
#start_time \t #value \t #duration \t #label

A typical call would then be:

`jams.util.import_lab('segment_open', names=range(4), usecols=[0, 2, 3], sep='\t')`

but I can't overwrite the behaviour that the duration is calculated.

So maybe we could add a flag in the function call to disable this?

bmcfee:
fixes #107

---
This change is [Reviewable](https://reviewable.io/reviews/marl/jams/125)

### DIFF
--- a/jams/tests/util_test.py
+++ b/jams/tests/util_test.py
@@ -15,20 +15,30 @@ import six
 
 
 def test_import_lab():
-
+    # Test a lab-file import
     labs = [r'''1.0 1
-3.0 2''', r'''1.0 2.0 a
-2.0 4.0 b''']
+                3.0 2''',
+            r'''1.0 1
+                3.0 2''',
+            r'''1.0 2.0 a
+                2.0 4.0 b''',
+            r'''1.0 1.0 c
+                2.0 2.0 d''']
 
-    intervals = [np.array([[1.0, 1.0], [3.0, 3.0]]),
+    intervals = [np.array([[1.0, 3.0], [3.0, 3.0]]),
+                 np.array([[1.0, 1.0], [3.0, 3.0]]),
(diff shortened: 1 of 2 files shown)


bmcfee:
Note: this PR changes the default behavior of `import_lab` on event annotations.

Previously, if there were only two columns on input, then the `duration` field was set to 0 as it was assumed to correspond to instantaneous events.

You can still achieve this behavior by setting `infer_duration=False` for this kind of input, but when the flag is set to `True` (default, to be consistent with interval inputs), duration is computed by the deltas between successive rows.  The final event is assumed to have zero duration.

This may cause hiccups in some data set import scripts.  However, I couldn't think of a better/more consistent way to parameterize this behavior.

bmcfee:
tagging @stefan-balke

stefan-balke:
Will check asap.

bmcfee:
Also tagging @urinieto since it might break some of his import scripts.

stefan-balke:
Tried it and LGTM!

bmcfee:
@urinieto can you take a peek at this and give me your :euro: e-2?

bmcfee:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/marl/jams/125)*

bmcfee:
@urinieto timed out!  But note that the salami importer will need to be updated.

PR #127, opened 2016-07-28
fixing jamsframe append type error

### LINKED ISSUE #126: Type inference error in annotation.append
This just bit me, and needs fixing:

``` python
n [1]: import jams

In [2]: an = jams.Annotation(namespace='tag_open')

In [3]: an.append(time=0, duration=1, value='foo')

In [4]: an.append(time=0, duration=2, value='75')

In [5]: an.data
Out[5]: 
    time  duration    value confidence
0 0 days  00:00:01      foo       None
1 0 days  00:00:02  0:00:00        NaT

In [6]: an.data.value
Out[6]: 
0        foo
1    0:00:00
Name: value, dtype: object

In [7]: an.data.value[1]
Out[7]: datetime.timedelta(0)
```

Note that the second row there was added as a string `"75"`, but pandas interpreted it as a `timedelta`.  This causes downstream validation errors, and is a no-good-very-bad thing.

bmcfee:
Trying to fix #126

---

This change is [Reviewable](https://reviewable.io/reviews/marl/jams/127)

### DIFF
--- a/jams/core.py
+++ b/jams/core.py
@@ -681,11 +681,13 @@ class JamsFrame(pd.DataFrame):
         if duration is None or not (duration >= 0.0):
             raise ParameterError('duration={} must be a non-negative number'.format(duration))
 
-        n = len(self)
-        self.loc[n] = {'time': pd.to_timedelta(time, unit='s'),
-                       'duration': pd.to_timedelta(duration, unit='s'),
-                       'value': value,
-                       'confidence': confidence}
+        time = pd.to_timedelta(time, unit='s')
+        duration = pd.to_timedelta(duration, unit='s')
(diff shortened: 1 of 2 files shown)


bmcfee:
Pushing upstream to https://github.com/pydata/pandas/issues/13829

bmcfee:
Confirmed to be a bug in pandas, not jams.  There's nothing to do in this PR, aside from the added test.

Since merging #123, this could be pretty painful until pandas rights itself.  Should we hold off on releasing the next jams until then?  Or not worry about it, and tell people it's a known issue?

I don't foresee it being a commonly encountered problem, though it could raise some serious problems when dealing with open tag data (eg, lastfm) that may have numeric-looking labels.

bmcfee:
The workaround in #150 resolves this.

Issue #145, opened 2017-03-19
Searching in JAMS with `None` and `str` values


urinieto:
I'm getting a `TypeError` when searching annotations within a JAMS file that contain a specific key that can either be a string or a `None`.

Here an example: 

when I do something this:
`jam.search(namespace="segment_open").search(**{"Sandbox.boundaries_id": "sf"})`

and some annotations inside `jam` have `boundaries_id=None` and others have `boundaries_id="sf"`, the `search` method raises a `TypeError`, like this:

```python
/home/uri/anaconda3/lib/python3.6/site-packages/jams/core.py in search(self, **kwargs)
   1340 
   1341         for annotation in self:
-> 1342             if annotation.search(**kwargs):
   1343                 results.append(annotation)
   1344 

/home/uri/anaconda3/lib/python3.6/site-packages/jams/core.py in search(self, **kwargs)
    449 
    450                 if isinstance(obj, JObject):
--> 451                     match |= obj.search(**r_query)
    452 
    453         return match

/home/uri/anaconda3/lib/python3.6/site-packages/jams/core.py in search(self, **kwargs)
    442         for key in r_query:
    443             if hasattr(self, key):
--> 444                 match |= match_query(getattr(self, key), r_query[key])
    445 
    446         if not match:

/home/uri/anaconda3/lib/python3.6/site-packages/jams/core.py in match_query(string, query)
   1836 
   1837     elif isinstance(query, six.string_types):
-> 1838         return re.match(query, string) is not None
   1839 
   1840     else:

/home/uri/anaconda3/lib/python3.6/re.py in match(pattern, string, flags)
    170     """Try to apply the pattern at the start of the string, returning
    171     a match object, or None if no match was found."""
--> 172     return _compile(pattern, flags).match(string)
    173 
    174 def fullmatch(pattern, string, flags=0):

TypeError: expected string or bytes-like object
```

Is there a way to circumvent this? Shouldn't `search` allow for this type of searches? Thanks!

bmcfee:
Hrm, not sure I'm following this entirely.  Can you gist a minimal jams object that triggers this behavior?

bmcfee:
Derp, nevermind, this does the trick:

```python
In [16]: J = jams.JAMS()

In [17]: ann = jams.Annotation(namespace='beat')

In [18]: ann.sandbox
Out[18]: <Sandbox: >

In [19]: ann.sandbox.foo = None

In [20]: J.annotations.append(ann)

[...]

PR #146, opened 2017-03-19
added type check in search re.match

bmcfee:
This PR fixes a type error in search when the target field is not a string but the query is.  Originally reported as #145.

New test case was added to cover this issue, so I think this is good to go.

---
This change is [Reviewable](https://reviewable.io/reviews/marl/jams/146)


urinieto:
Great, looking good!

Issue #93, opened 2016-01-28
Pretty printing: _repr_html_ and _repr_svg_


bmcfee:
It would be nice if we could have better pretty-printing of jams objects (and annotations, specifically) for notebook environments, ie, by implementing a `_repr_html_()` method.

justinsalamon:
is there no existing code for pretty printing json files in general that we can leverage?

bmcfee:
Sure, but I'm thinking something more akin to [msaf's plots](https://github.com/urinieto/msaf/blob/master/examples/Run%20MSAF.ipynb) that render out a table for the annotation objects. Doing ``display(jam.annotations[i].data)`` is nice and all to render the data frame object, but the information could be encoded much more concisely by exploiting the regular structure of jamsframes.

A really cool version of this would combine all annotations into a single table-like widget, where each row corresponds to an annotation.  This way, you could easily get a visual digest of all annotations in a compact space.

One can also imagine having different rendering helpers for different namespace types, so that events draw like events and regions draw like regions.  (Even fancier, pitches and chords draw like piano roll, etc.)

justinsalamon:
>One can also imagine having different rendering helpers for different namespace types, so that events draw like events and regions draw like regions. (Even fancier, pitches and chords draw like piano roll, etc.)

So basically a jupyter version of sonic visualiser :P Yes, that would be cool.

bmcfee:
> So basically a jupyter version of sonic visualiser :P

Yup!  But it could also be used outside of notebook environments, eg, for a web-based visualizer.

justinsalamon:
>eg, for a web-based visualizer.

I think I can see where you're going with this... and I like it.

bmcfee:
Thinking a bit more about this, it might be tricky to properly handle overlapping intervals, eg:

| time | duration | value | confidence |
| --- | --- | --- | --- |
| 0 | 5 | guitar | 1 |
| 2.5 | 2.5 | vocals | 1 |

should render as two rows that overlap by half horizontally, but 

| time | duration | value | confidence |
| --- | --- | --- | --- |
| 0 | 2.5 | guitar | 1 |
| 2.5 | 5 | vocals | 1 |

should render as one row.

Ideally, a css layout engine should do this for us, but I'm not convinced that it will work.

justinsalamon:
Why should the second example render as one row? One could argue that each
instrument should have a separate track (row) throughout the visualization.
That's how sequencers do it at least.

[...]

PR #149, opened 2017-05-09
[CR needed] alternate AnnotationData backend

bmcfee:
This is  work in progress, but implements an alternate backend for annotation data.

TODO:

- [x] trim
- [x] slice
- [x] deep serialization (eg for vector format)
- [x] pretty printing
- [x] purge all pandas references
- [x] settle on final object hierarchy / naming conventions
- [x] update converter scripts and documentation / examples

### DIFF
--- a/jams/core.py
+++ b/jams/core.py
@@ -26,7 +26,6 @@ Object reference
     AnnotationMetadata
     Curator
     Annotation
-    AnnotationData
     Observation
     Sandbox
     JObject
@@ -56,7 +55,7 @@ __all__ = ['load',
            'JObject', 'Sandbox',
            'Annotation', 'Curator', 'AnnotationMetadata',
            'FileMetadata', 'AnnotationArray', 'JAMS',
-           'AnnotationData', 'Observation']
+           'Observation']
 
 
 @contextlib.contextmanager
@@ -504,197 +503,6 @@ Observation = namedtuple('Observation',
 '''Core observation type: (time, duration, value, confidence).'''
 
 
-class AnnotationData(object):
-
-    __dense = False
-
-    def __init__(self):
-        self.obs = SortedListWithKey(key=self._key)
-
-    @classmethod
-    def _key(cls, obs):
-        return obs.time
-
-    @property
-    def dense(self):
-        '''Boolean to determine whether the encoding is dense or sparse.
-
-        Returns
-        -------
-        dense : bool
-            `True` if the data should be encoded densely
-            `False` otherwise
-        '''
-        return self.__dense
-
-    @dense.setter
-    def dense(self, value):
-        '''Setter for dense'''
-        self.__dense = value
-
-    def add_observation(self, time=None, duration=None, value=None,
-                        confidence=None):
-        '''Add a single observation.
-
-        Parameters
-        ----------
-        time : number >= 0, required
-            The start time of this observation
-
-        duration : number >= 0, required
-            The duration of this observation
-
-        value
-            The value for the observation
-
-        confidence
-            The confidence for the observation
-        '''
-        self.obs.add(Observation(time=time, duration=duration,
-                                 value=value, confidence=confidence))
-
-    def append_records(self, records):
-        '''Add observations from row-major storage.
-
-        This is primarily useful for deserializing sparsely packed data.
-
-        Parameters
-        ----------
-        records : iterable of dicts or Observations
-            Each element of `records` corresponds to one observation.
-        '''
-        for obs in records:
-            if isinstance(obs, Observation):
-                self.add_observation(**obs._asdict())
-            else:
-                self.add_observation(**obs)
-
-    def append_columns(self, columns):
-        '''Add observations from column-major storage.
-
-        This is primarily used for deserializing densely packed data.
-
-        Parameters
-        ----------
-        columns : dict of lists
-            Keys must be `time, duration, value, confidence`,
-            and each much be a list of equal length.
-
-        '''
-        self.append_records([dict(time=t, duration=d, value=v, confidence=c)
-                             for (t, d, v, c)
-                             in six.moves.zip(columns['time'],
-                                              columns['duration'],
-                                              columns['value'],
-                                              columns['confidence'])])
-
-    def to_interval_values(self):
-        '''Extract observation data in a `mir_eval`-friendly format.
-
-        Returns
-        -------
-        intervals : np.ndarray [shape=(n, 2), dtype=float]
-            Start- and end-times of all valued intervals
-
-            `intervals[i, :] = [time[i], time[i] + duration[i]]`
-
-        labels : list
-            List view of value field.
-        '''
-        ints, vals = [], []
-        for obs in self.obs:
-            ints.append([obs.time, obs.time + obs.duration])
-            vals.append(obs.value)
-
-        return np.array(ints), vals
-
-    def to_dataframe(self):
-        '''Convert this annotation to a pandas dataframe.
-
-        Returns
-        -------
-        df : pd.DataFrame
-            Columns are `time, duration, value, confidence`.
-            Each row is an observation.
-        '''
-        return pd.DataFrame.from_records(list(self.obs),
-                                         columns=['time', 'duration',
-                                                  'value', 'confidence'])
-
-    @property
-    def __json__(self):
-        '''JSON encoding attribute'''
-
-        if self.dense:
-            times, durations, values, confidences = [], [], [], []
-            for (t, d, v, c) in self.obs:
-                times.append(t)
-                durations.append(d)
-                values.append(serialize_obj(v))
-                confidences.append(c)
-
-            return dict(time=times,
-                        duration=durations,
-                        value=values,
-                        confidence=confidences)
-        else:
-            return [dict(time=o.time,
-                         duration=o.duration,
-                         value=serialize_obj(o.value),
-                         confidence=o.confidence) for o in self.obs]
-
-    def __len__(self):
-        return len(self.obs)
-
-    def __eq__(self, other):
-        return (isinstance(other, self.__class__) and
-                self.obs == other.obs)
-
-    def __repr__(self):
-        return '<{}: {:d} observations>'.format(self.__class__.__name__,
-                                                len(self))
-
-    def __iter__(self):
-        return iter(self.obs)
-
-    def to_html(self):
-        '''Render this annotation list in HTML
-
-        Returns
-        -------
-        rendered : str
-            An HTML table containing this annotation's data.
-        '''
-        out = r'''<table border="1" class="dataframe">
-                    <thead>
-                        <tr style="text-align: right;">
-                            <th></th>
-                            <th>time</th>
-                            <th>duration</th>
-                            <th>value</th>
-                            <th>confidence</th>
-                        </tr>
-                    </thead>'''
-        out += r'''<tbody>'''
-        for i, o in enumerate(self.obs):
-            out += r'''<tr>
-                            <th>{:d}</th>
-                            <td>{:0.6f}</td>
-                            <td>{:0.6f}</td>
-                            <td>{:}</td>
-                            <td>{:}</td>
-                        </tr>'''.format(i,
-                                        o.time,
-                                        o.duration,
-                                        o.value,
-                                        o.confidence)
-        out += r'''</tbody></table>'''
-        return out
-
-    def _repr_html_(self):
-        return self.to_html()
-
-
 class Annotation(JObject):
     """Annotation base class."""
 
@@ -736,16 +544,13 @@ class Annotation(JObject):
 
         self.namespace = namespace
 
-        self.data = AnnotationData()
-
-        # Set the data export coding to match the namespace
-        self.data.dense = schema.is_dense(self.namespace)
+        self.data = SortedListWithKey(key=self._key)
 
         if data is not None:
             if isinstance(data, dict):
-                self.data.append_columns(data)
+                self.append_columns(data)
             else:
-                self.data.append_records(data)
+                self.append_records(data)
 
         if sandbox is None:
             sandbox = Sandbox()
@@ -755,7 +560,7 @@ class Annotation(JObject):
         self.time = time
         self.duration = duration
 
-    def append(self, **kwargs):
+    def append(self, time=None, duration=None, value=None, confidence=None):
         '''Append an observation to the data field
 
         Parameters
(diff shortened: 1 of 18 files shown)


bmcfee:
Okay, this PR is now functional in that all tests pass again.  So if you never touched a `JamsFrame` object directly in application code (and there are few reasons to do so), then everything should just work. 
 That said, the design is up for discussion though, particularly seeking input from @justinsalamon and @ejhumphrey.

The new implementation replaces the `JamsFrame` object by an `AnnotationData` object, which at the schema level covers both `DenseObservation` and `SparseObservationList`, though not using `JObject`s because I don't think we should have sandbox support in the observation list.

Under the hood, `AnnotationData` encapsulates the observations with an ordered (by `time`) list of sparse observations.  Observations are implemented as named tuples `(time, duration, value, confidence)`, again not using JObject so that we can preserve slots and field ordering.

This namedtuple implementation also makes individual observations immutable, which I see as a Good Thing. If you want to modify observations, it's best to make a new annotation object and re-insert them after modification.  Note that this is the reverse of how the `DataFrame` implementation worked: the individual observations there were mutable (it's easy to modify entries in-place), but the container itself was much more rigid (adding and deleting observations was difficult).  I think the mutable-container/immutable-observation model makes a lot more sense for the common cases (e.g., adding observations to an existing annotation).

You can iterate over observations in the usual way:
```python
for obs in ann.data:
    print(obs.time, obs.duration, obs.value, obs.confidence)
```

@justinsalamon and I kicked around the idea of having an `export-to-dataframe` method, which should be pretty trivial to cook up.  Similarly, we might want to consider adding other mir_eval-style converters, such as `to_event_values` for instantaneous observations.

Pretty-printing is the last major hurdle, but I think we can get over that fairly easily.

jams/core.py
```diff
@@ -496,49 +498,21 @@ class Sandbox(JObject):
...
-        '''Construct a new JamsFrame object.
-
-        Parameters
-        ----------
-        data
-            Optional data for the new JamsFrame, in any format supported
-            by `pandas.DataFrame.__init__`.
 
-            Fields must be `['time', 'duration', 'value', 'confidence']`.
+class AnnotationData(object):
```
ejhumphrey: honest question: does inheriting directly from `SortedListWithKey` buy us anything? This object owning an `obs` object feels like an unnecessary abstraction, but I'm not sure if this would result in a weird interface.

thoughts?

jams/core.py
```diff
@@ -704,23 +569,48 @@ def to_interval_values(self):
...
-        return jf
+    def __eq__(self, other):
+        return (isinstance(other, self.__class__) and
+                self.obs == other.obs)
+
+    def __repr__(self):
+        return '<{}: {:d} observations>'.format(self.__class__.__name__,
+                                                len(self))
+
+    def __iter__(self):
```
ejhumphrey: for example, we wouldn't need to implement this method, or L601

ejhumphrey: or L604, i think?

jams/core.py
```diff
@@ -557,139 +531,30 @@ def dense(self, value):
...
-        '''
-        # Encode time properly
-        frame.time = pd.to_timedelta(frame.time, unit='s')
-        frame.duration = pd.to_timedelta(frame.duration, unit='s')
-
-        # Properly order the columns
-        frame = frame[cls.fields()]
+    def add_observation(self, time=None, duration=None, value=None,
+                        confidence=None):
+        idx = self.obs.bisect_key(time)
```
ejhumphrey: if nothing else, I think we can use `SortedListWithKey` directly ([docs](http://www.grantjenks.com/docs/sortedcontainers/sortedlistwithkey.html#sortedcontainers.SortedListWithKey.L.add)). I poked around, and calling `add` sorts items in-line, so `self.obs.add(Observation(...))
` should work.

additionally, i think if AnnotationData is a sublcass of `SortedListWithKey`, then we might subclass `add` directly? and maybe call it `ObservationArray`? idk?

jams/core.py
```diff
@@ -704,23 +569,48 @@ def to_interval_values(self):
...
+                durations.append(d)
+                values.append(serialize_obj(v))
+                confidences.append(c)
+
+            return dict(time=times,
+                        duration=durations,
+                        value=values,
+                        confidence=confidences)
+        else:
+            return [dict(time=o.time,
```
ejhumphrey: if `Observation` is an object rather than a NamedTuple, might we get some of this for free?

jams/core.py
```diff
@@ -1016,10 +892,10 @@ def trim(self, start_time, end_time, strict=False):
         # We do this rather than copying and directly manipulating the
         # annotation' data frame (which might be faster) since this way trim is
         # independent of the internal data representation.
-        for idx, obs in self.data.iterrows():
+        for obs in self.data.obs:
```
ejhumphrey: I *would* be keen to simplify this and make the object directly iterable, by a subclass or whatever

ejhumphrey: oh wait, it is. I'd remove `.obs` to dig in the preferred interaction paradigm

jams/nsconvert.py
```diff
@@ -137,12 +140,38 @@ def can_convert(annotation, target_namespace):
     return False
 
 
+def pop_data(annotation):
```
ejhumphrey: feels like this should be a method of `AnnotationData`?

ejhumphrey:
well, first off, 💯  for a PR / RFC that deletes more code than it adds.

I do actually like that this gets us away from pandas (maybe as an optional dependency?), but we only trade dependencies. Is the development velocity on `sortedcontainers` lower than pandas, and that's why it's a win?

question: didn't pandas provide some neat wins for muda? shifting annotations / observations in time and whatnot?

jams/core.py
bmcfee: At this point, I'm thinking it's a bad idea to extend from other peoples' classes because it makes the API encapsulation boundary porous.

jams/core.py
bmcfee: I tried that.  Maybe I did it wrong, but it doesn't allow you to insert items in a way that violates sort order.

jams/core.py
bmcfee: Some of what?  I mean, we can always do `dict(obs._asdict())`, but that doesn't solve the dict-of-lists/list-of-dicts problem resulting from sparse/dense observation lists.

jams/nsconvert.py
bmcfee: Yeah, I just had that thought too.  Will migrate in a bit.

jams/core.py
ejhumphrey: ya, you can't [insert](http://www.grantjenks.com/docs/sortedcontainers/sortedlistwithkey.html#sortedcontainers.SortedListWithKey.L.insert) or [append](http://www.grantjenks.com/docs/sortedcontainers/sortedlistwithkey.html#sortedcontainers.SortedListWithKey.L.append) if it violates order, but add should sort.

jams/core.py
ejhumphrey: "some of what": the serialization waltz, but yea, sparse vs dense makes this tricky.

bmcfee:
> I do actually like that this gets us away from pandas (maybe as an optional dependency?), but we only trade dependencies. Is the development velocity on sortedcontainers lower than pandas, and that's why it's a win?

I think we still need pandas as a dependency for loading lab/csv files (utils, import scripts, etc).  But it's good to get the hacky dataframe dependencies out from under the jams object hierarchy.

Sortedcontainers is a much simpler project, and seems reasonably mature at this point.

> question: didn't pandas provide some neat wins for muda? shifting annotations / observations in time and whatnot?

It did, but that's not a huge deal.  It's better IMO to make observations immutable and put a little more friction into annotation manipulation.  Muda will need a little retrofitting, but it's probably about an afternoon's worth of work.

jams/core.py
ejhumphrey: fair enough ... I guess JamsFrame inherited from DataFrame, but maybe that was half the issue?

if we're not going to inherit from it, I'd propose making obs some level of private (dunders?) so that users are discouraged from using it as part of the API. However, I don't like having to wrap a bunch of pre-existing functionality ... SortedListWithKey is quite fully loaded.

Also worth mentioning here: What gets added to an AnnotationData object? `Observation`s or the fields of an observation? or, said differently, who is responsible for creating the Observation object, the user or the AnnotationData object? because, if it's the former, I think that's an argument for inheritance. 

related pop quiz: in Jams, who creates an Annotation when adding it to a JAMS object, the user or the owning object? (because we should follow that paradigm for consistency)

jams/core.py
bmcfee: Are you talking about `Annotation` or `AnnotationData`?

The latter is iterable (wrapping the sorted list); the former is not and wasn't previously.  We could make it iterable, but that sounds like a change better suited to a separate PR.

jams/core.py
bmcfee: > if we're not going to inherit from it, I'd propose making obs some level of private (dunders?) so that users are discouraged from using it as part of the API.

Yeah, I'd be down with that.  I'd say sunder not dunder, ie `data._obs`.



> Also worth mentioning here: What gets added to an AnnotationData object? Observations or the fields of an observation? or, said differently, who is responsible for creating the Observation object, the user or the AnnotationData object? because, if it's the former, I think that's an argument for inheritance.

It depends: if you're adding a bunch of records at once, it's typically in deserialized form (ie after a json parse), so it'd be list-of-dict/dict-of-list in.

The core add routine takes the fields for a single observation record and packs them in an `Observation` for adding.

I don't see what this has to do with inheritance though?



> related pop quiz: in Jams, who creates an Annotation when adding it to a JAMS object, the user or the owning object? (because we should follow that paradigm for consistency)

Should we?  I see the `AnnotationData` class as relatively private, because it diverges somewhat from the schema. `Annotation` provides an interface for adding observations, and that's all a user should need.

bmcfee:
Note: coverage is down here because I added html rendering for `AnnotationData`.  I don't think unit tests are critical on that function.

justinsalamon:
> Note: coverage is down here because I added html rendering for AnnotationData.

Would it be a good/bad idea, since we still have the pandas dependency anyhow, to do viz by exporting to a dataframe and then using the pandas built in viz? This could be nice in that you could use the variety of viz options offered by the pandas API for dataframes (e.g. peak(), head(), etc.)

bmcfee:
> Would it be a good/bad idea, since we still have the pandas dependency anyhow, to do viz by exporting to a dataframe and then using the pandas built in viz?

I don't think we should rely on it, but it's always an option.  The dataframe converter is done anyway, so you can do
```python
df = ann.data.to_dataframe()
df.head(10)
```
etc.

jams/core.py
bmcfee: Thinking more about this:

Another option is to drop `AnnotationData` entirely, keep `ann.data` as a sorted container object, and bubble up the rest of the functionality for adding observations, iterations, etc to `Annotation`.  This would be a more radical change in the API than I'd originally intended, but it seems like maybe a good idea provided we're willing to break stuff anyway.

bmcfee:
I thought some more about this PR last night, and I think I'm convinced that we should drop the separate `AnnotationData` class, and promote most of its functionality up to the `AnnotationData` level.  The `data` field will then be a `SortedListWithKey` object, and the `Annotation` object will handle all (de)serialization and management of observations.

As noted in the review threads, this would break some API for folks that expect the `ann.data` object to behave in certain ways, but we're already doing that anyway.  The result should be a slightly simpler API, eg:
```python
ann.to_interval_values()
```
instead of
```python
ann.data.to_interval_values()
```
etc.

Is there any reason not to do this within this PR?

justinsalamon:
This will likely break more things in libs that build on jams, but as you say if we break things might as well go all in if it's worth it.

If the data structure is completely hidden and only accesible via the API, we should make sure the API supports operations for e.g. checking if the annotation is empty.

The one possible caveat to this change might be that since an annotation also has other fields (metada, time, duration, sandbox), the lack of an explicit data object could get a tad confusing in certain cases, e.g. len(ann)=? print(ann)=? etc

bmcfee:
> If the data structure is completely hidden and only accesible via the API, we should make sure the API supports operations for e.g. checking if the annotation is empty.

Well, the `data` field is specified in the schema as a list/array (actually, either a denseobservationlist or a sparseobservationlist), so we shouldn't make it entirely private.





> The one possible caveat to this change might be that since an annotation also has other fields (metada, time, duration, sandbox), the lack of an explicit data object could get a tad confusing in certain cases, e.g. len(ann)=? print(ann)=? etc

I think it's reasonable to treat `Annotation` as a [collection](https://docs.python.org/3.6/library/collections.abc.html#collections.abc.Collection) of observations with some metadata on the side. Then we can define `len(ann)` as the number of observations, `iter(ann)` iterates over observations, and `__contains__` tests if an observation belongs to the list (not that we really need that, but it doesn't hurt).

`repr(ann)` currently returns a JObject-style repr string like `<Annotation: data, annotation_metadata, ...>`, so we wouldn't necessarily need to change anything there.  I wouldn't mind changing JObject's repr format overall, since it's not terribly informative for distinguishing between different objects of the same tyep, but that's a separate issue.

bmcfee:
Amending my previous statement: I don't think we can consistently make `Annotation` implement the Collection ABC because it would change the semantics inherited from `JObject` for `__len__` and `__contains__`, which both apply to the object's `__dict__`.  OTOH there's no problem in juts implementing `__iter__`.

bmcfee:
@ejhumphrey I think this one's stable for eyeballs again.

Coverage drop is primarily due to html rendering not being covered, but I'd prefer to push that to #93 and implement it across the board.

jams/nsconvert.py
```diff
@@ -159,7 +165,13 @@ def note_midi_to_hz(annotation):
     '''Convert a pitch_midi annotation to pitch_hz'''
 
     annotation.namespace = 'note_hz'
-    annotation.data.value = 440 * (2.0 ** ((annotation.data.value - 69.0)/12.0))
+    data = annotation.pop_data()
```
ejhumphrey: i can neither articulate why nor propose something "better", but the semantics of this feel weird for some reason.

tests/namespace_tests.py
```diff
@@ -30,12 +29,10 @@ def test_ns_time_invalid():
...
         ann = Annotation(namespace='onset')
 
-        # Bypass the safety chceks in add_observation
-        ann.data.loc[0] = {'time': pd.to_timedelta(data['time'], unit='s'),
-                           'duration': pd.to_timedelta(data['duration'],
-                                                       unit='s'),
-                           'value': None,
-                           'confdence': None}
+        # Bypass the safety checks in append
+        ann.data.add(Observation(time=data['time'],
```
ejhumphrey: this looks *so* much nicer

ejhumphrey:
Couple thoughts, minor apologies if / when relevance is a stretch:
- Not concerned with the coverage drop; does coveralls have a setting that says "don't worry about drops if the total coverage is over a certain level" (say, 98?)
- It has since occurred to me that all of my other projects using jams have things like `for idx, obs in ann.data.iterrows()` ... is this a common pattern (or am I weird); and if it is common, are we throwing deprecation warnings for this usage? or is it gone outright?

jams/nsconvert.py
bmcfee: Do you object to the operation itself, or just the naming?

bmcfee:
> Not concerned with the coverage drop; does coveralls have a setting that says "don't worry about drops if the total coverage is over a certain level" (say, 98?)

I'd prefer to keep the strict coverage check in place, just so that we're aware of it at all times.



> It has since occurred to me that all of my other projects using jams have things like for idx, obs in ann.data.iterrows() ... is this a common pattern (or am I weird); and if it is common, are we throwing deprecation warnings for this usage? or is it gone outright?

Gone outright: any previous access to the `data` field is not guaranteed to work in general.  We throw deprecation warnings for some things in 0.2.3, but it was too much of a mess to do it in full generality without also flagging internal access that was going to change anyway.

The `idx, obs` pattern can be recuvered by
```python
for idx, obs in enumerate(ann):
    ...
```
though there's almost never a good reason to want the index itself.

bmcfee:
Hashed out in person with @ejhumphrey  -- this one's good to go.

Issue #147, opened 2017-05-09
pandas 0.20 compatibility


bmcfee:
This will be fun.  Not sure how deep this all goes, but I think we're in for some compatibility headaches.

```python
In [3]: J = jams.load('trapped.jams')
/home/bmcfee/git/jams/jams/core.py:1852: FutureWarning: pandas.tslib is deprecated and will be removed in a future version.
You can access Timedelta as pandas.Timedelta
  if isinstance(obj, pd.tslib.Timedelta):
```

bmcfee:
For reference, here are the [release notes](http://pandas.pydata.org/pandas-docs/stable/release.html#pandas-0-20-0-0-20-1).

Running locally, it's a total disaster:
```
Ran 759 tests in 23.096s

FAILED (failures=271)
```

justinsalamon:
😱😱😱

bmcfee:
It looks like these all(?) stem from the same issue inside `JamsFrame.add_observation` that is not allowing us to append strings into the value column.

bmcfee:
Punting upstream to https://github.com/pandas-dev/pandas/issues/16309 , but I don't think we should wait on pandas to fix it for us.

Issue #122, opened 2016-07-20
Redo tests in py.test


bmcfee:
nosetests is deprecated, py.test is the new hotness.  This will be an easy, but time-consuming job.

PR #150, opened 2017-05-10
Pandas 020 compatibility

bmcfee:
This PR is a short-term fix for #147, intended for immediate release.  It should get the current jams running on pandas 0.20.

### DIFF
--- a/jams/nsconvert.py
+++ b/jams/nsconvert.py
@@ -24,6 +24,7 @@ __CONVERSION__ = defaultdict(defaultdict)
 
 __all__ = ['convert', 'can_convert']
 
+
 def _conversion(target, source):
     '''A decorator to register namespace conversions.
 
@@ -167,7 +168,9 @@ def note_hz_to_midi(annotation):
     '''Convert a pitch_hz annotation to pitch_midi'''
 
     annotation.namespace = 'note_midi'
-    annotation.data.value = 12 * (np.log2(annotation.data.value) - np.log2(440.0)) + 69
+
+    logf = np.log2(np.asarray(annotation.data.value.values, dtype=np.float))
+    annotation.data.value = 12 * (logf - np.log2(440.0)) + 69
     return annotation
 
 
@@ -185,7 +188,8 @@ def pitch_hz_to_midi(annotation):
     '''Convert a pitch_hz annotation to pitch_midi'''
 
     annotation.namespace = 'pitch_midi'
-    annotation.data.value = 12 * (np.log2(annotation.data.value) - np.log2(440.0)) + 69
+    logf = np.log2(np.asarray(annotation.data.value.values, dtype=np.float))
+    annotation.data.value = 12 * (logf - np.log2(440.0)) + 69
     return annotation
 
 
--- a/tests/namespace_tests.py
+++ b/tests/namespace_tests.py
@@ -230,7 +230,11 @@ def test_ns_note_hz_invalid():
 
     # note: 1j should also be invalid, but currently not caught
     # by the schema validation and hence removed from the test
-    for value in ['a', -23, None, True]:
+    # for value in ['a', -23, None, True]:
+
+    # None and True get auto-cast to float now, so we can't rely
+    # on validation here
+    for value in ['a', -23]:
         yield __test, value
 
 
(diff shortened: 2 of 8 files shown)


bmcfee:
Note: the type inference bug in pandas is really biting us hard here.  Column data types for values and confidences are now left as `object`, which might break some upstream code.  I don't know of a good way to fix this, since value type inference depends on the namespace, which is one level up from `JamsFrame`.

justinsalamon:
hero.

bmcfee:
Okay, tests should be passing at this point.  I had to do some pretty gnarly things to get this working.

It's not exactly bit-for-bit compatible with the previous release for two reasons:

1. Column type inference is somehow different now, and pandas is leaving many things as `object`
2. Reindexing is also different now, so some of the slicing and dicing that happens, eg, in `trim` results in strange behavior.

Neither of these things should affect most common behaviors, and the loose guarantees made at the API level are still satisfied.

However, in-place modification of jamsframes is severely broken, eg, if you want to apply a numpy `ufunc` to a numerically-typed value field.  The better way to do this is to iterate over observations, since that will be compatible going forward.

Also, `JamsFrame.equal` is no longer reliable as an equivalence test because it's too strict about indexing and dtype matching.  In the unit tests where this comes up, I replaced the call to `equals` by pandas' internal `assert_frame_equal` function, which allows you to loosen some of the checks (index type, dtype matching, etc).

My recommendation is to merge this as soon as possible, deprecate the `JamsFrame` object, put out a 0.2.3 release ASAP and then turn around quickly for 0.3, where we can put some of the safety checks back in place.

*EDIT*: by the time I'd finished writing this, the tests had failed on 0.19 because pandas moved the `testing` submodule.  You gotta be kidding.

bmcfee:
I think this one's good to go, provided we can tolerate the shortcomings listed above. Anyone care to cr?

bmcfee:
I'm gonna take a stab at schema-derived type inference for value/confidence columns, just to see if we can't smooth over the transition a little more.

bmcfee:
I'm holding off on merging this until I can sort out why muda tests are still failing, but otherwise it's good for review.

bmcfee:
It appears that something has gone wrong within `deepcopy`, and objects are sharing when they shouldn't be.

bmcfee:
Okay, muda build passes again without modification.  The bug was indeed due to deepcopy not actually going deep enough.  I can't wait to ditch this pandas backend.

Meanwhile, this is ready for CR: @ejhumphrey interested?

bmcfee:
FWIW scaper tests all pass locally with pandas 0.20 as well, so @justinsalamon can relax.

bmcfee:
msaf master also passes tests on this PR.  I tagged @urinieto and @ejhumphrey for a quick CR: both of you don't need to do this, but it would be nice to get some extra eyes on this hack before merging.

It's all stop-gap in this PR just to keep things running until we can merge #149 and get on with the next major revision without pandas backing.

jams/nsconvert.py
```diff
@@ -167,7 +168,9 @@ def note_hz_to_midi(annotation):
     '''Convert a pitch_hz annotation to pitch_midi'''
 
     annotation.namespace = 'note_midi'
-    annotation.data.value = 12 * (np.log2(annotation.data.value) - np.log2(440.0)) + 69
+
+    logf = np.log2(np.asarray(annotation.data.value.values, dtype=np.float))
```
ejhumphrey: nit: any value to breaking this pitch-hz mapping out as a separate function (in the spirit of do one thing)

tests/namespace_tests.py
```diff
@@ -322,9 +334,13 @@ def __test(value):
 
     # note: 1j should also be invalid, but currently not caught
     # by the schema validation and hence removed from the test
-    for value in ['a', None, True]:
+    # for value in ['a', None, True]:
+
+    # None and True cast to float
+    for value in ['a']:
         yield __test, value
```
ejhumphrey: aren't yield fixtures deprecated now? or am i confused?

bmcfee: in pytest, but jams is not yet switched over to pytest.  We'll handle that in #122

jams/nsconvert.py
bmcfee: eh, it's all getting redone in #149 anyway, so I'm not gonna sweat it

bmcfee:
Thanks @ejhumphrey !

Issue #152, opened 2017-05-13
Relocate interval_values converter, add deprecation warnings


bmcfee:
If we're removing JamsFrame, we should give folks a chance to adapt.

`to_interval_values` will move up to the `Annotation` level, so we can put a deprecation warning in place.

We should also put an observation iterator up in `Annotation` that yields `Observation` namedtuples.  This way, downstream projects can adopt the upcoming api without breaking things in the short-term.

PR #153, opened 2017-05-13
0.2.3 -> 0.3.0 shims

bmcfee:
This PR adds deprecation warnings and some forward-looking compatibility for #149 .

Implements #152.

### DIFF
--- a/jams/core.py
+++ b/jams/core.py
@@ -29,7 +29,7 @@ Object reference
     JamsFrame
     Sandbox
     JObject
-
+    Observation
 """
 
 import json
@@ -45,6 +45,7 @@ import contextlib
 import gzip
 import copy
 import sys
+from collections import namedtuple
 
 from decorator import decorator
 
@@ -57,7 +58,8 @@ from .exceptions import JamsError, SchemaError, ParameterError
 __all__ = ['load',
            'JObject', 'Sandbox', 'JamsFrame',
            'Annotation', 'Curator', 'AnnotationMetadata',
-           'FileMetadata', 'AnnotationArray', 'JAMS']
+           'FileMetadata', 'AnnotationArray', 'JAMS',
+           'Observation']
 
 
 def deprecated(version, version_removed):
@@ -513,6 +515,11 @@ class JObject(object):
             valid = False
 
         return valid
+
+
+Observation = namedtuple('Observation',
+                         ['time', 'duration', 'value', 'confidence'])
+'''Core observation type: (time, duration, value, confidence).'''
 
 
 class Sandbox(JObject):
@@ -1213,6 +1220,15 @@ class Annotation(JObject):
         duration = timedelta_to_float(self.data.duration.values)
 
         return np.vstack([times, times + duration]).T, list(self.data.value)
+
+    def __iter_obs__(self):
+        for _, (t, d, v, c) in self.data.iterrows():
+            yield Observation(time=t.total_seconds(),
+                              duration=d.total_seconds(),
+                              value=v, confidence=c)
+
+    def __iter__(self):
+        return self.__iter_obs__()
 
 
 class Curator(JObject):
--- a/jams/core.py
+++ b/jams/core.py
@@ -46,6 +46,9 @@ import gzip
 import copy
 import sys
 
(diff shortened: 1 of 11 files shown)


jams/core.py
```diff
@@ -1168,6 +1195,25 @@ def slice(self, start_time, end_time, strict=False):
...
+            `intervals[i, :] = [time[i], time[i] + duration[i]]`
+
+        labels : list
+            List view of value field.
+        '''
+
+        times = timedelta_to_float(self.data.time.values)
+        duration = timedelta_to_float(self.data.duration.values)
+
+        return np.vstack([times, times + duration]).T, list(self.data.value)
```
ejhumphrey: why `np.vstack` over `np.array`..?

jams/eval.py
```diff
@@ -104,8 +104,8 @@ def beat(ref, est, **kwargs):
     namespace = 'beat'
     ref = coerce_annotation(ref, namespace)
     est = coerce_annotation(est, namespace)
-    ref_interval, _ = ref.data.to_interval_values()
-    est_interval, _ = est.data.to_interval_values()
+    ref_interval, _ = ref.to_interval_values()
```
ejhumphrey: ohgeez this is so much cleaner.

tests/jams_test.py
```diff
@@ -211,10 +211,16 @@ def test_jamsframe_interval_values():
...
     jf = jams.JamsFrame.from_dataframe(df)
 
-    intervals, values = jf.to_interval_values()
+    warnings.resetwarnings()
+    warnings.simplefilter('always')
+    with warnings.catch_warnings(record=True) as out:
+        intervals, values = jf.to_interval_values()
+        assert len(out) > 0
+        assert out[0].category is DeprecationWarning
+        assert 'deprecated' in str(out[0].message).lower()
```
ejhumphrey: nice

jams/core.py
bmcfee: to concatenate them along the second axis; historical style holdover from the days before arbitary concat support.

but this is getting wiped in #149 also.

jams/core.py
```diff
@@ -1168,6 +1202,34 @@ def slice(self, start_time, end_time, strict=False):
...
+
+    def __iter_obs__(self):
+        for _, (t, d, v, c) in self.data.iterrows():
```
ejhumphrey: I find it odd that iterating `self` is (or will be) a shortcut to `self.data.obs` ... which feels like two unnecessary levels of abstraction? right?

bmcfee: `self.data.obs` is now (will be) `self.data`

bmcfee: ... but that's neither here nor there.  The point of this change is to add forward support for:

```python
for obs in annotation:
   # process jams.Observation object
```

the under-the-hood parts of how that iterator works won't matter.

ejhumphrey: I guess I meant that `self -> Annotation`, `self.data -> AnnotationData`, `self.data.obs -> SortedListWithKey`, and so only the last contains the items yielded by the iterator.

ejhumphrey: ah yes, i realize I forgot to contextualize my comment -- this change is fine; however, it occurs to me that the data structure is maybe deeper than necessary? not necessarily for iteration, but mutating it?

bmcfee: Right -- check the changes I made to #149 last night / this morning.  That's all cleaned up now (and either way, irrelevant to this PR).

ejhumphrey: ah yes, i see now that my understanding of things elsewhere is stale.

bmcfee:
Aside from deprecation of `to_interval_values` and adding the `Observation` type and annotation iterator, is there anything else that we can do here to make the transition easier?

bmcfee:
.. I'm inclined to think there isn't much else we can do here, since any other deprecation warnings would apply to methods called internally as well as the public.

Unless there are objections from @ejhumphrey , I think this is ready to merge and push as 0.2.3 release.

ejhumphrey:
i have no other thoughts or opinions at the moment, tho I apologize in advance for when they occur to me in 4 days.

PR #155, opened 2017-05-19
accelerated pitch contour sonification. fixes #151

### LINKED ISSUE #151: improve efficiency of pitch contour sonification
The sonifier for this namespace is woefully slow because it does multiple passes over the data.

It could be made much more efficient by first constructing an index of the contours in one pass rather than relying on post-filtering for each contour id.

I'll hold off on this until after #149 merges to avoid multiple rewrites.

bmcfee:
This PR implements #151 and should accelerate the sonification of pitch contours.

Benchmarks will be pasted below when they're finished.

### DIFF
--- a/jams/sonify.py
+++ b/jams/sonify.py
@@ -11,7 +11,7 @@ Sonification
 '''
 
 from itertools import product
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 import six
 import numpy as np
 import mir_eval.sonify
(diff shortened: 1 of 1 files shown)


bmcfee:
First cut appears to not provide any improvement, but I think this is because my benchmark only has a single contour.  Consequently, the pain-point of multiple iterations over the data is not the key bottleneck.

@rabitt do you have any multi-contour annotations kicking around that I can benchmark on?

bmcfee:
I faked up some data by duping contours in medleydb.  Sonifying musicdelta - reggae, pitch_hz 1 with a second contour at f*1.5 from the first, the current implementation takes

```
CPU times: user 32 ms, sys: 0 ns, total: 32 ms
Wall time: 32.9 ms
```

the new implementation takes
```
CPU times: user 28 ms, sys: 8 ms, total: 36 ms
Wall time: 39.2 ms
```

This is vastly overshadowed by the time spent within the `validator` when doing automatic namespace conversion, so that's probably a better place to focus attention.  Unless there are objections, I'll merge this and close it out.

PR #158, opened 2017-05-26
improved repr and repr_html

bmcfee:
Tagging @ejhumphrey @justinsalamon @rabitt @urinieto 

This PR is looking at improvements to our repr strings, implementing #93.  Once that's hammered out, i'll start looking at HTML formatting for the notebook.

So far, here's what we've got:

```python
>>> J
<JAMS(annotations=[4 annotations],
      file_metadata=<FileMetadata(...)>,
      sandbox=<Sandbox(...)>)>

>>> J.file_metadata
<FileMetadata(artist='The Beatles',
              duration=119.333,
              identifiers=<Sandbox(...)>,
              jams_version='0.2.0',
              release='',
              title='11_-_Do_You_Want_To_Know_A_Secret')>

>>> J.annotations
[4 annotations]

>>> J.annotations[0]
<Annotation(annotation_metadata=<AnnotationMetadata(...)>,
            data=<114 observations>,
            duration=None,
            namespace='chord',
            sandbox=<Sandbox(...)>,
            time=0)>
```

I've added a couple of helper methods and an external `summary` function that knows how to make condensed summaries of JObjects.

Note: the old version, for comparison's sake, would look like this:
```python
>>> J
<JAMS: sandbox, file_metadata, annotations>

>>> J.file_metadata
<FileMetadata: title, identifiers, artist, jams_version, duration, release>

>>> J.annotations
[<Annotation: namespace, sandbox, data, time, duration, annotation_metadata>,
 <Annotation: namespace, sandbox, data, time, duration, annotation_metadata>,
 <Annotation: namespace, sandbox, data, time, duration, annotation_metadata>,
 <Annotation: namespace, sandbox, data, time, duration, annotation_metadata>]

>>> J.annotations[0]
<Annotation: namespace, sandbox, data, time, duration, annotation_metadata>
```

What do folks think?  Is there anything that seems too complex, or other information that should be surfaced in the repr?  This still doesn't quite satisfy the intended semantics of `repr` as returning a string which reproduces the same object when entered into the interpreter, but it's at least more informative than the old version.

### DIFF
--- a/jams/core.py
+++ b/jams/core.py
@@ -321,6 +321,19 @@ class JObject(object):
                            for p in props)
         return '<{}({:})>'.format(self.type, params)
 
+    def _repr_html_(self):
+        schema = self.__schema__
+        props = []
+        if schema:
+            props = sorted(schema['properties'].keys())
+
+        out = '<div><dl class="dl-horizontal">'
+        for prop in props:
+            out += '<dt>{}</dt>'.format(prop)
+            out += '<dd>{}</dd>'.format(summary_html(self[prop]))
+        out += '</dl></div>'
+        return out
+
     def __summary__(self):
         return '<{}(...)>'.format(self.type)
 
@@ -1044,7 +1057,7 @@ class Annotation(JObject):
     def __iter__(self):
         return iter(self.data)
 
-    def to_html(self, max_rows=None):
+    def _to_html(self, max_rows=None):
         '''Render this annotation list in HTML
 
         Returns
@@ -1085,6 +1098,76 @@ class Annotation(JObject):
         out += r'''</tbody>'''
 
         out += r'''</table>'''
+        return out
+
+    def to_html(self, max_rows=None):
+        '''Render this annotation list in HTML
+
+        Returns
+        -------
+        rendered : str
+            An HTML table containing this annotation's data.
+        '''
+        n = len(self.data)
+
+        div_id = _get_divid(self)
+
+        out = r'''
+                    <div class="panel panel-default">
+                        <div class="panel-heading" role="tab" id="heading-{0}">
+                                <a  role="button"
+                                    data-toggle="collapse"
+                                    data-parent="#accordion"
+                                    href="#{0}"
+                                    aria-expanded="false"
+                                    class="collapsed label label-primary"
+                                    aria-controls="{0}">
+                                <em>{1:s}</em>
+                                <span class="badge pull-right">{2:d}</span>
+                                </a>
+                        </div>'''.format(div_id, self.namespace, n)
+
+        out += r'''     <div id="{0}" class="panel-collapse collapse"
+                             role="tabpanel" aria-labelledby="heading-{0}">
+                            <div class="panel-body">'''.format(div_id)
+
+        # -- Annotation content starts here
+        out += r'''<table border="1" class="dataframe">
+                    <thead>
+                        <tr style="text-align: right;">
+                            <th></th>
+                            <th>time</th>
+                            <th>duration</th>
+                            <th>value</th>
+                            <th>confidence</th>
+                        </tr>
+                    </thead>'''.format(self.namespace, n)
+
+        out += r'''<tbody>'''
+
+        if max_rows is None or n <= max_rows:
+            out += self._fmt_rows(0, n)
+        else:
+            out += self._fmt_rows(0, max_rows//2)
+            out += r'''<tr>
+                            <th>...</th>
+                            <td>...</td>
+                            <td>...</td>
+                            <td>...</td>
+                            <td>...</td>
+                        </tr>'''
+            out += self._fmt_rows(n-max_rows//2, n)
+
+        out += r'''</tbody>'''
+
+        out += r'''</table>'''
+
+        # Maybe we should wrap the data, metadata, and sandbox in tabs
+        # or in divs?
+        out += r'''<div>{}</div>'''.format(self.annotation_metadata._repr_html_())
+        out += r'''<div>{}</div>'''.format(self.sandbox._repr_html_())
+
+        out += r'''</div></div></div>'''
         return out
 
     def _fmt_rows(self, start, end):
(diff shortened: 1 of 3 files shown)


bmcfee:
Quick update:

I bootstrapped this up a little, and it's starting to look better.  Repr'ing a jams object looks like:

image

Each annotation can be drop-down expanded like:
image

I'm still not crazy about the generic JObject formatting, like in the file_metadata/sandbox display in the first screencap.  Anyone have suggestions for how that could be improved?  @justinsalamon @rabitt ?

bmcfee:
Last round of updates I think.

We now have display property ordering and formatting, and a slightly cleaner layout.  Missing values are indicated in red.

At the top level, a jams object renders something like
image

Drilling down into an annotation, we get
image

For unstructured objects (JObject, Sandbox), properties are displayed with their key name and sorted alphabetically.  Other classes have things ordered in a more reasonable way, eg:

```python
>>> print(repr(J.annotations[0]))
<Annotation(namespace='chord',
            time=0,
            duration=None,
            annotation_metadata=<AnnotationMetadata(...)>,
            data=<36 observations>,
            sandbox=<Sandbox(...)>)>
```

I think the spacing of the observation /ann.metadata fields could be improved, but otherwise it seems pretty usable to me.  Anyone have differing opinions / ideas for improvement?

bmcfee:
Test coverage is up, and things are reasonably pretty now.  Or at least not aesthetically corrosive.

Anyone want to CR and/or pull and test-drive?

bmcfee:
Back in the real world today.  CR / tire-kicking: @justinsalamon ? @ejhumphrey ?

bmcfee:
Note: this is the only major hurdle to getting a pre-release of 0.3 out, so I'd like to get this done sooner rather than later.

jams/core.py
```diff
@@ -309,7 +309,83 @@ def __len__(self):
...
+            content = summary_html(self[prop])
+
+            prop_class = 'default'
+            if not content:
+                prop_class = 'danger'
+
+            out += '<div class="panel panel-{}">'.format(prop_class)
+
+            if (isinstance(self[prop], (JObject, AnnotationArray, dict))
+               and content):
```
ejhumphrey: sorry for the nit, but does this lint? i'd expect an extra whitespace in front of the `and`?

bmcfee: yeah it's ugly, but my linter likes it better than the correct way, for some reason.

bmcfee: and yes, it passes pycodestyle and pylint only complains about redefining eval and some import shenanigans.

ejhumphrey: 🤔

jams/core.py
```diff
@@ -1048,26 +1163,50 @@ def to_html(self):
...
+        return out
+
+    def _fmt_rows(self, start, end):
+        out = ''
+        for i, obs in enumerate(self.data[start:end], start):
             out += r'''<tr>
                             <th>{:d}</th>
-                            <td>{:0.6f}</td>
-                            <td>{:0.6f}</td>
+                            <td>{:0.3f}</td>
```
ejhumphrey: is there any way to hijack float precision from, say, numpy? since this change is fixing the precision, but as a private method, I don't see a sane / clean way to adjust the behavior. `jams.set_precision`?

ejhumphrey: derp, i meant `jams.set_printoptions()`

ejhumphrey: also IIRC @cjacoby knows a sweet defaults library?

bmcfee: or a more general `jams.set_printoptions` ala pandas?  Maybe.  I guess we could add a global config dict for this sort of thing, but I wouldn't mind punting that downstream for a point release.

It would probably be best to add a rc/config file as well, which opens up a few more headaches than I'd like to deal with right now.

ejhumphrey: it seems that we agree.

tests/test_jams.py
```diff
@@ -101,8 +101,21 @@ def test_jobject_nonzero(data, value):
     assert J.__nonzero__() == value
 
 
-# Sandbox
+def test_jobject_repr():
+    repr(jams.JObject(foo=1, bar=2))
```
ejhumphrey: should this verify that `repr` returns something substantial?

tests/test_jams.py
```diff
@@ -101,8 +101,21 @@ def test_jobject_nonzero(data, value):
...
+
+
+def test_jobject_repr_html():
+    # Test once with empty
+    J2 = jams.JObject()
+    J2._repr_html_()
 
+    # And once with some nested values
+    J = jams.JObject(foo=1, bar=dict(baz=3), qux=[1], quux=None)
+    J._repr_html_()
```
ejhumphrey: assert that this returns something? Ideally we'd want to validate that it's well-formatted HTML, but I couldn't easily (sub 30 seconds) find a library that does this?

tests/test_jams.py
bmcfee: hah, probably.  Actually none of these tests check output, just that the code executes properly.  I guess we could put some dummy checks in.

tests/test_jams.py
```diff
@@ -519,6 +532,18 @@ def test_jobject_bad_field():
     jam.out_of_schema = None
 
 
+def test_jams_repr(input_jam):
```
ejhumphrey: same thoughts on this block as above

tests/test_jams.py
bmcfee: I looked for a while, and couldn't find any usable libraries for this purpose.

ejhumphrey:
overall this looks great! i'm curious if there's hope for HTML validation. like i said in the comments, i'm surprised and not optimistic after some brief googling, but I'd feel much better knowing that we're producing valid HTML. everything else is a nit.

tests/test_jams.py
bmcfee: discussed offline -- since it seems to work offline, and the functionality isn't critical to anything downstream, we're okay with not being 100% thorough on validating the output of the repr/repr_html functions.

bmcfee:
So is this one good to merge then?

ejhumphrey:
lgtm

PR #160, opened 2017-07-05
Numpy JSON serialization

### LINKED ISSUE #159: NumPy float format serialization issues
To be clear, what follows is *arguably* user error, but more graceful failure would have saved me some time (and I'm getting different behavior on different environments, which didn't help). 

I made the mistake of populating an annotation with np.float32 values. You can too with the following (note that this isn't what I did, but is functionally equivalent).

```python
>>> ann = jams.Annotation(namespace='tag_open', time=0, duration=2.0)
>>> data = np.array([[1.0, 1.0, 0.5]], dtype=np.float32)
>>> for time, duration, conf in data:
    ann.data.add_observation(
            time=time, duration=duration, value='foo',
            confidence=conf)
```

Cool, no obvious problems yet.
```python
# I look mostly fine.
>>> print(ann.__json__)  

{'data': [{'duration': 1.0, 'confidence': 0.5, 'value': 'foo', 'time': 1.0}], 'duration': 2.0, 
 'annotation_metadata': {'version': '', 'corpus': '', 'annotator': {}, 'annotation_rules': '',
 'annotation_tools': '', 'data_source': '', 'curator': {'name': '', 'email': ''}, 'validation': ''}, 
 'namespace': 'tag_open', 'sandbox': {}, 'time': 0}
```

However, these are (silently) `np.float32`s, though they don't quite look it (other numbers that don't have the same representation in single / double precision floats have a better tell), and generally won't in tracebacks.

```python
# I fail :o(
>>> jam = jams.JAMS(annotations=[ann]).save("deleteme.jams", strict=False)  

<snip json traceback>
TypeError: 0.5 is not JSON serializable
```

I'm wondering if the API could help cast non-standard numerical types here? perhaps as a converter dict like in pandas? or perhaps just over data arrays? If default, we'd also want to warn on the casting behavior.

Or, at the bare minimum, it'd be helpful to throw a more informative warning like "hey you have nonserializable types! they are X and they're in field Y. Go back and look at what you're putting in the data structure"

thoughts?

bmcfee:
This PR fixes #159 by doing explicit type coercion on numpy scalars when converting to JSON.

It's a bit of a stop-gap solution for now.  I think we should at some point re-structure how `serialize_obj` is implemented and used, so that it applies transparently throughout all `__json__` calls.  That's probably out of scope for this fix though, so this PR should be ready to go.

### DIFF
--- a/jams/core.py
+++ b/jams/core.py
@@ -273,7 +273,7 @@ class JObject(object):
             if hasattr(item, '__json__'):
                 filtered_dict[k] = item.__json__
             else:
-                filtered_dict[k] = item
+                filtered_dict[k] = serialize_obj(item)
 
         return filtered_dict
 
@@ -693,8 +693,8 @@ class Annotation(JObject):
         >>> ann.append(time=3, duration=2, value='E#')
         '''
 
-        self.data.add(Observation(time=time,
-                                  duration=duration,
+        self.data.add(Observation(time=float(time),
+                                  duration=float(duration),
                                   value=value,
                                   confidence=confidence))
 
@@ -1992,7 +1992,13 @@ def serialize_obj(obj):
 
     '''
 
-    if isinstance(obj, np.ndarray):
+    if isinstance(obj, np.integer):
+        return int(obj)
+
+    elif isinstance(obj, np.floating):
+        return float(obj)
+
+    elif isinstance(obj, np.ndarray):
         return obj.tolist()
 
     elif isinstance(obj, list):
--- a/tests/test_jams.py
+++ b/tests/test_jams.py
@@ -1117,3 +1117,9 @@ def test_deprecated():
 
         # And that it says the right thing (roughly)
         assert 'deprecated' in str(out[0].message).lower()
+
+
+def test_numpy_serialize():
+
+    jobj = jams.JObject(key=np.float32(1.0))
+    jobj.__json__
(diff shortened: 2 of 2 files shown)


jams/core.py
```diff
@@ -693,8 +693,8 @@ def append(self, time=None, duration=None, value=None, confidence=None):
         >>> ann.append(time=3, duration=2, value='E#')
         '''
 
-        self.data.add(Observation(time=time,
-                                  duration=duration,
+        self.data.add(Observation(time=float(time),
+                                  duration=float(duration),
                                   value=value,
                                   confidence=confidence))
```
ejhumphrey: I know `confidence` can be None, but might we want a similar coercion here too?

tests/test_jams.py
```diff
@@ -1117,3 +1117,9 @@ def _foo():
 
         # And that it says the right thing (roughly)
         assert 'deprecated' in str(out[0].message).lower()
+
+
+def test_numpy_serialize():
```
ejhumphrey: This looks good for `JObject`, which I guess is testing `serialize_object` under the hood ... I'm wondering if it's worth checking the float casting for the observation data in `Annotation`?

jams/core.py
bmcfee: Confidence is not typed in the schema. Technically it can be anything, just like value, so we can't coerce here.

ejhumphrey: reeeheheally... 🤔

bmcfee: Yup.  The reasoning here is confidence is namespace-dependent; sometimes it's required (tempo), sometimes it's optional (msd_tagtraum).

tests/test_jams.py
bmcfee: Exactly.  I just pushed up a second test for Annotation, working on observation fields.  I think that should cover it for now.

ejhumphrey:
LGTM 👍

Issue #132, opened 2016-09-07
Support multiple JSON Schema Validator Backends?


ejhumphrey:
As implicitly acknowledged in #46, validating against the JSON schema is expensive, slow, and when running things at scale, really making me think about skipping it, which is a red flag. If I'm thinking about bailing on validation, others aren't far behind me. 

I did a little digging, and it seems this is more a function of `jsonschema` being slow than schema validation being inherently slow. For example, this: https://github.com/epoberezkin/ajv#performance

Couple thoughts:
- I recognize that these are not Python libraries, but it demonstrates the variability in validator performance.
- The fastest one has a [command line interface](https://github.com/jessedc/ajv-cli) that we could consider wrapping. This wouldn't (and shouldn't) be the default validator, but multiple backends could allow for more flexible validator support.
- There don't actually seem to be many "native" options for JSON schema validation in Python, so wrapping one in another language might be simplest way to speed this up.

thoughts? complete heresy?

bmcfee:
So, are you suggesting to make a python wrapper on ajv?  That sounds useful, especially if its api could look like jsonschema (or at least, close enough to be swappable).  Alternately, if it was done via some kind of javascript/node bridge (does this exist?) then it could persist some state so it's not rebuilding the whole validator every time it gets called.

I'd be open to using such a library in JAMS, but it would definitely complicate the dependency chain to have non-python dependencies.  OTOH, if the hypothesized pyajv included the full ajv source (which seems legit, given the MIT license on ajv), it could work.

ejhumphrey:
> So, are you suggesting to make a python wrapper on ajv?

yeah, exactly.

> Alternately, if it was done via some kind of javascript/node bridge (does this exist?) then it could persist some state so it's not rebuilding the whole validator every time it gets called.

yeeaaaa I have no idea. Previous experience attempting to bridge scripting languages was ... not great ... so wrapping a CLI might be the easier out.

> OTOH, if the hypothesized pyajv included the full ajv source (which seems legit, given the MIT license on ajv), it could work.

Agreed. I'm curious if either of the `ajv` or `ajv-cli` projects would be keen to take on a python wrapper _in_ the project (probably the latter, if at all).

bmcfee:
> so wrapping a CLI might be the easier out.

Sure... doesn't the cli require node (or something similar) though? I mean, the thing's written in js.



> 'm curious if either of the ajv or ajv-cli projects would be keen to take on a python wrapper in the project (probably the latter, if at all).

[...]

Issue #46, opened 2015-07-08
Validation caching


bmcfee:
It would be nice if we could somehow checksum and cache expensive operations, such as schema validation.

Once we validate a jams object, we shouldn't need to validate it again if it hasn't changed.

urinieto:
Is there an easy way to do that using the @cache decorator?

bmcfee:
> Is there an easy way to do that using the @cache decorator?

Yes, but we'd do better to use joblib.Memory directly.

It's still not a great solution though, since it would cache the entire object.  I'd rather checksum (or otherwise hash) the object first, and then cache that.

bmcfee:
Having read through joblib.Memory docs, I'm not sure this idea is really feasible.  The big hurdle here is that the object being validated against the schema is split across multiple objects (eg, each annotation, sandbox, etc. is a separate object) that all get down-cast to dicts for the purposes of validation.  This means that caching at the top-level needs to maintain cache consistency of all of its sub-objects, which is going to require a lot of overhead.  I doubt it will buy us much.

I'll close this out, but if anyone wants to take it up again, feel free to re-open.

PR #170, opened 2017-09-06
[CR] Accelerating schema validation

bmcfee:
This PR accelerates schema validation, in particular for dense array data, following the thread over in #132 / #46 .

So far, the main improvement is due to vectorizing schema validation at the annotation level, rather than validating each observation independently.  This provides somewhere between 4x and 8x speedup without changing any functionality.

### DIFF
--- a/jams/core.py
+++ b/jams/core.py
@@ -581,7 +581,7 @@ class JObject(object):
         valid = True
 
         try:
-            jsonschema.validate(self.__json__, schema.JAMS_SCHEMA)
+            jsonschema.validate(self.__json__, self.__schema__)
 
         except jsonschema.ValidationError as invalid:
             if strict:
@@ -760,12 +760,15 @@ class Annotation(JObject):
         JObject.validate
         '''
 
-        valid = super(Annotation, self).validate(strict=strict)
-
         # Get the schema for this annotation
         ann_schema = schema.namespace_array(self.namespace)
 
+        valid = True
+
         try:
+            jsonschema.validate(self.__json_light__(data=False),
+                                schema.JAMS_SCHEMA)
+
             # validate each record in the frame
             data_ser = [serialize_obj(obs) for obs in self.data]
             jsonschema.validate(data_ser, ann_schema)
@@ -1210,6 +1213,9 @@ class Annotation(JObject):
 
     @property
     def __json__(self):
+        return self.__json_light__(data=True)
+
+    def __json_light__(self, data=True):
         r"""Return the JObject as a set of native data types for serialization.
 
         Note: attributes beginning with underscores are suppressed.
@@ -1220,7 +1226,10 @@ class Annotation(JObject):
             if k.startswith('_'):
                 continue
             elif k == 'data':
-                filtered_dict[k] = self.__json_data__
+                if data:
+                    filtered_dict[k] = self.__json_data__
+                else:
+                    filtered_dict[k] = []
 
             elif hasattr(item, '__json__'):
                 filtered_dict[k] = item.__json__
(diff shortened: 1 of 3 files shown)


bmcfee:
I think there's also redundant serialization happening within `validate`, and that's now the majority of computation time in validation: more than half the time is now spent in serialization, not validation.

The top-level JAMS validator [here](https://github.com/marl/jams/blob/master/jams/core.py#L1751-L1756) does the following:

1. Validate the entire JAMS object and its sub-objects according to the JAMS schema.
2. If (1) passes, validate each annotation independently according to namespace schema.

Within the annotation validator [here](https://github.com/marl/jams/blob/master/jams/core.py#L763-L780), the following happens:

3. Run the generic validator for Annotations
4. Run the namespace-dependent observation array validator

What this all boils down to is that observations get serialized three times: once for the top-level pass (1), once for the generic Annotation pass (3) and once for the namespace pass (4).

So: how can we cut down on redundant serialization?

1. Skip annotation serialization in `JAMS.validate`, since those will be done independently anyway.  We'd still need to do a post-check to make sure that the `annotations` array is well formed (ie, an AnnotationArray type), but that's relatively cheap.

2. Hack `Annotation.validate` to bypass observations in the first pass.  Alternately, promote the vectorized namespace validator to a full annotation validator, and skip the `super` validate [here](https://github.com/marl/jams/blob/master/jams/core.py#L763).  (I think I like this second option better.)

@ejhumphrey what do you think?

ejhumphrey:
I think I need time to digest this later, but first a question – are these mutually exclusive options?

bmcfee:
> are these mutually exclusive options?

Nope, mostly just interested in feedback on whether there might be alternative routes to eliminating redundant serialization (short of caching), and what the best way to go about it might be.

ejhumphrey:
first read, both seem good, and I think doing first (2) then (1) if needed  ... makes sense to me?

bmcfee:
Ok, doing a light-weight serialization of `Annotation` (skipping the `data` field) shaves ~1s off my running example (52s -> 7.1s -> 6.2s).

Gonna try the light-weight ser on `JAMS` now.

bmcfee:
Bypassing the annotation array validation within `JAMS.validate()` brings the running example down to 3.31s.  This seems acceptably fast to me, given where it started out.

bmcfee:
Rolled in a bugfix for #171 that I noticed while implementing tests on the accelerated validation code.

I think this one's ready for CR.

jams/core.py
```diff
@@ -760,15 +760,18 @@ def validate(self, strict=True):
...
+        valid = True
 
         try:
+            jsonschema.validate(self.__json_light__(data=False),
+                                schema.JAMS_SCHEMA)
+
             # validate each record in the frame
-            for rec in self.data:
-                jsonschema.validate(serialize_obj(rec), ann_schema)
+            data_ser = [serialize_obj(obs) for obs in self.data]
```
rabitt: Is this where the speed up is coming from?

bmcfee: That's the first speedup.  The second speedup comes from eliminating redundant serialization calls, as noted above in-thread.

bmcfee:
One more benchmark, using the biggest file in the MedleyDB dataset: `AlexanderRoss_VelvetCurtain.jams` (24MB).

* Load with no validation:
```python
In [4]: %time J = jams.load('AlexanderRoss_VelvetCurtain.jams', validate=False)
CPU times: user 1.01 s, sys: 48 ms, total: 1.06 s
Wall time: 1.05 s
```
* Load with old, slow validation:
```python
In [5]: %time J = jams.load('AlexanderRoss_VelvetCurtain.jams', validate=True)
CPU times: user 1min 35s, sys: 72 ms, total: 1min 35s
Wall time: 1min 35s
```
* Load with new, fast validation:
```python
In [6]: %time J = jams.load('AlexanderRoss_VelvetCurtain.jams', validate=True)
CPU times: user 6.63 s, sys: 24 ms, total: 6.66 s
Wall time: 6.66 s
```

Issue #109, opened 2016-04-08
jams.sample() [ie sampling a series of discrete annotation elements into a time-series]


justinsalamon:
In automatic transcription, it is not uncommon to evaluate the estimated note sequence not just using the note transcription metrics, but also using the melody extraction metrics. Since the melody extraction metrics are frame-based, this requires sampling the note sequence on a fixed time grid to convert it to a time series in the expected format.

Are there any other tasks/use cases where this type of series sampling would come in handy? Is this something worth implementing as a utility function in JAMS? For non-overlapping events the mapping is straight forward. For potentially overlapping events it's a little less, but (for example) the `value` array could contain a list (or tuple?) in each frame containing all values mapped to the same timestamp.

Thoughts?

bmcfee:
I think what you're asking for is something like mir_eval's `intervals_to_samples` function, but for overlapping data.  That is, something that provides regularly spaced samples of the interval-labeled data.

I've implemented this kind of thing a few times, and yeah, it would be useful to have a generic solution.  The tricky part here is, like you say, to handle overlapping intervals sanely, and make it generic enough to apply to different namespaces.

I think this can be done with some judicious use of lambda functions.  For example, you could imagine providing one function that maps a `value` field to an object, and a second function that combines (reduces) objects to a single object.

For example, the mapper could be scikit-learn `LabelEncoder.transform` method, which knows how to translate strings into sparse encoding vectors, and the `reduce` function could be `np.maximum` (or any ufunc).  The semantics here would be that each frame gets the union of labels corresponding to intervals containing said frame.

More concretely, it might look like the following:
```python
>>> LE = sklearn.preprocessing.LabelEncoder()
>>> LE.fit(array_of_tags)
>>> frames, labels = jams.sample_annotation(annotation, map=LE.transform, reduce=np.maximum)
```

Does this make any kind of sense?  Am I over-engineering?

justinsalamon:
It does make sense. In my specific case I'd define custom map and reduce functions (I just want to keep the pitch value in Hz, and since notes don't overlap reduce would just pass the value along), but I like this logic.

What did you have in mind for the returned `frames` and `labels`? Just `np.ndarray`s with timestamps and values?

bmcfee:
> What did you have in mind for the returned frames and labels? Just np.ndarrays with timestamps and values?

- `frames` would be an ndarray of time indices (in seconds)
[...]

PR #173, opened 2017-09-09
[CR] annotation sampler

bmcfee:
This PR implements #109


bmcfee:
This one's ready for CR, @rabitt or @justinsalamon want to chime in?

The scope of the PR is just implementing value sampling for a set of specified time instants.  It does not implement any automatic time gridding.  This decision is mainly because the `duration` field is optional for an annotation, so samples generated from multiple annotations for the same track might not be easy to align.  What are yalls thoughts on that?

PR #177, opened 2017-10-28
display viz mappings: equal parts RFC and PR

ejhumphrey:
The jams.display submodule rocks but I've had to override the `VIZ_MAPPING` symbol table to use it with different namespaces, e.g. `tag_open`. I've taken the liberty of adding that one here, and alphabetizing the symbol table so we can add to it with some kind of convention going forward.

but!

this might be indicative of something we want to fix in general, and I'm curious to kick it around (here, I guess?). Should a namespace JSON file *also* specify how it should be visualized? seems like adding this info in two places is a design smell.

also: happy to move this to an issue first.

### DIFF
--- a/jams/display.py
+++ b/jams/display.py
@@ -132,14 +132,15 @@ def piano_roll(annotation, **kwargs):
 
 VIZ_MAPPING = OrderedDict()
 
-VIZ_MAPPING['segment_open'] = intervals
+VIZ_MAPPING['beat'] = event
+VIZ_MAPPING['beat_position'] = beat_position
 VIZ_MAPPING['chord'] = intervals
 VIZ_MAPPING['multi_segment'] = hierarchy
-VIZ_MAPPING['pitch_contour'] = pitch_contour
-VIZ_MAPPING['beat_position'] = beat_position
-VIZ_MAPPING['beat'] = event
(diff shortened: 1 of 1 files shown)


bmcfee:
> I've taken the liberty of adding that one here, and alphabetizing the symbol table so we can add to it with some kind of convention going forward.

I don't think it should be alphabetically or lexically ordered.  The choice of `OrderedDict` was intentional, and meant to reflect priority and specificity of the different annotation viz functions when multiple matches are possible. For instance, basically everything is convertible to `tag_open`, but you wouldn't want to use that viz for things where more precise displays are possible.

I agree that this is not well documented though, so we should improve on that.

bmcfee:
(ping @ejhumphrey)  any thoughts on above?

bmcfee:
@ejhumphrey can we close this one?

bmcfee:
Reupping this one again: @ejhumphrey can you revert the ordering changes (with a comment to indicate that order reflects precedence of namespace coercion)?  Otherwise, I think `tag_open` should be a low-priority viz, so its placement at the end is fine.

bmcfee:
Chatted with @ejhumphrey out of band.  We'll close this out and reopen as a new PR just with the added viz link.

PR #191, opened 2018-06-01
Add schemata for scaper and urbansound

justinsalamon:


justinsalamon:
@bmcfee I think this is ready for CR

PR #194, opened 2018-11-15
Create jams_to_mirex_pattern.py

irisyupingren:
A converter between the jams JKUPDD namespace and the submission format for the MIREX task

### DIFF
--- a/scripts/jams_to_mirex_pattern.py
+++ b/scripts/jams_to_mirex_pattern.py
@@ -1,65 +1,89 @@
 
+#!/usr/bin/env python
+'''Convert JAMS to MIREX pattern discovery format
+
+Example usage:
+
+./jams_to_mirex_pattern.py /path/to/file.jams outputname
+'''
+
 import csv
 import json
+import sys
+from argparse import ArgumentParser
+import os
+import jams
 
-x = open('/path/to/file.jams').read()
-x = json.loads(x)
 
-record = []
 
-for x in x["annotations"]:
-    # looping through the annotators
-    metadata = x["annotation_metadata"]
-    namespace = x["namespace"]
-    
-    # annotator names
-    anno = metadata["annotator"]
-    
-    # actual data of annotation
-    d = x["data"]
+def parse_arguments(args):
 
-    # open file to write to the MIREX pattern discovery format 
-    text_file = open("anno{0}.txt".format(anno), "w")
-    text_file.write("Pattern1"+"\n")
-    text_file.write("Occ1"+"\n")
-    pcount = 1 # keep pattern count
-    ocount = 1 # keep pattern occurrence count
+    parser = ArgumentParser(description='Parse JAMS annotations into the MIREX Pattern Discovery task format')
 
-    pastpid = 1 # initiatise the pattern count to compare for a change
-    pastoid = 1 # initiatise the occurrence count to compare for a change
-    
-    initpid = d[0]["value"]["pattern_id"]
+    parser.add_argument('infile', type=str, help='Input JAMS file')
+    parser.add_argument('output_prefix', type=str,
+                        help='Prefix for output lab files')
 
-    for y in d:
-        # lopping through the events given an annotator
-        time = y["time"]
-        dur = y["duration"]
+    return vars(parser.parse_args(args))
 
-        pid = y["value"]["pattern_id"]
+def run(infile='', output_prefix='annotation'):
+    with open(infile, 'r') as fdesc:
+        x = json.load(fdesc)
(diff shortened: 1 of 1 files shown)


bmcfee:
Apologies for the failing test build -- I'll get this patched up ASAP.

bmcfee:
@irisyupingren if you pull the latest master, and do `git rebase master` from this branch, the tests should start working again.

However, since this PR doesn't touch the core library, the test failures here aren't particularly relevant -- I'll get a CR out ASAP.  Thanks for doing this!

scripts/jams_to_mirex_pattern.py
```diff
@@ -0,0 +1,65 @@
+
```
bmcfee: There should be a shebang and docstring comment explaining what this does and how to use it:
```
#!/usr/bin/env python
'''Convert JAMS to MIREX pattern discovery format

Example usage:

./jams_to_mirex_pattern.py /path/to/jams.py /path/to/mirex_output.csv
'''

or something to that effect.

File: scripts/jams_to_mirex_pattern.py [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+import csv
+import json
+
+x = open('/path/to/file.jams').read()
```
bmcfee: A couple of points here:

1) It's better to use context managers for this, so that errors are handled properly.  Instead of

```python
x = open('file').read()
x = json.load(x)
```
do
```python
with open('file.jams', 'r') as fdesc:
    x = json.load(fdesc)
```

Though, really it's better to use `jams.load('file.jams')` so that the data structure is parsed into a JAMS object, rather than loading raw json.

2) Filenames should be given as command line arguments.  Check out the `jams_to_lab.py` script to see how we do this currently.

scripts/jams_to_mirex_pattern.py
```diff
@@ -0,0 +1,65 @@
...
+    namespace = x["namespace"]
+    
+    # annotator names
+    anno = metadata["annotator"]
+    
+    # actual data of annotation
+    d = x["data"]
+
+    # open file to write to the MIREX pattern discovery format 
+    text_file = open("anno{0}.txt".format(anno), "w")
```
bmcfee: again, context managers will help handle errors properly:

```python
with open('anno{0}.txt'.format(anno), 'w') as text_file:
    ...
```

scripts/jams_to_mirex_pattern.py
```diff
@@ -0,0 +1,65 @@
+
+import csv
+import json
+
+x = open('/path/to/file.jams').read()
+x = json.loads(x)
+
+record = []
+
+for x in x["annotations"]:
```
bmcfee: It's best to not re-use variable names like this: `x` takes on at least 3 different meanings in the current code, making it difficult to read.  It's better to do `for annotation in x['annotations']`.

Additionally, if you use `jams.load()` instead of raw json, you can use the jams search logic to narrow down to only the annotations in pattern discovery namespace:

```
jam = jams.load('file.jams')
for annotation in jam.annotations['pattern']:
    output each annotation...
```

scripts/jams_to_mirex_pattern.py
irisyupingren: Added argparse. 

Trying to load using jams.load but got a type error. Opened an issue. #196

scripts/jams_to_mirex_pattern.py
irisyupingren: yes that was dangerous sorry. 
same issue about using jams.load #196

scripts/jams_to_mirex_pattern.py
```diff
@@ -0,0 +1,89 @@
...
+    parser.add_argument('output_prefix', type=str,
+                        help='Prefix for output lab files')
+
+    return vars(parser.parse_args(args))
+
+def run(infile='', output_prefix='annotation'):
+    with open(infile, 'r') as fdesc:
+        x = json.load(fdesc)
+
+    record = []
```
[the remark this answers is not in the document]
bmcfee: You can get this stuff automatically by running pyflakes, flake8, or pylint on the script file.

bmcfee:
Hey @irisyupingren , sorry to take literally forever to get around to this again.  github wasn't showing your updated code in the review panel!  At any rate, this looks good now.  Is it okay to merge this for the next release?  Thanks again for doing this, and for your patience!

irisyupingren:
No worries! It's been a while and it's good that everything's still up to date. Sure, looking forward to the next release :D

Issue #92, opened 2015-12-14
Schema refactor


bmcfee:
Rehashing #40 after a conversation with @ejhumphrey 

There are good arguments for splitting the JAMS schema into smaller pieces that can be shared and repurposed.  Specifically, a database (eg, a mongodb key-value store) for managing jams collections could be more reasonable structured (and easily searchable) if the database contains individual annotation objects (indexed by track id) rather than full JAMS objects.

I propose that we refactor the jams schema so that annotations can exist independently of the JAMS file format.  Of course, the JAMS file format will still use annotation definitions, so there should be no observable difference in the way JAMS files work*; put another way, the API for JAMS files stays the same, and all the changes would be under the hood.

Digging in a bit more, the current schema looks like:
```
jams_schema
`- JAMS
   `- FileMetadata
   |  `- [more stuff]
   `- Annotations
   |  `- [more stuff]
   `- Sandbox
```
and the refactored schema might look like:
```
jams_common
`- Sandbox

jams_annotation
`- Annotations
   `- [more stuff]

jams_metadata
`- FileMetadata
   `- [more stuff]

jams_file
`- JAMS
   `- jams_metadata.FileMetadata
   `- jams_annotation.Annotations
   `- jams_common.Sandbox
```

What do folks think?

To make this happen, we'd have to get a better handle on json-schema inheritance, but I think it's totally possible.

* We might have to tweak the schema id's, which might require a slight modification to the spec.  Not sure about this yet.

ejhumphrey:
More related to this than worth spawning a new issue: I'd like to revisit / upvote a conversation about how identifiers / named entities are referenced in JAMS. For example, I'd like to tag a single annotation as being produced by some unique identifier, such that I can search a collection for all annotations performed by the same entity (human or algorithm). We've got the annotator dict, but it's a little too unconstrained to encourage any convention.

bmcfee:
I'm not sure that fits under the scope of JAMS per se; remember the headaches about filenames in #5? We eventually decided that that's better handled at the application level -- for better or worse.  I suspect that indexing annotation sources will have similar difficulties.

OTOH, if we do want to add support for foreign-key indexing (for tracks, annotators, etc), maybe it's worth reopening that discussion?

urinieto:
Could we simply add a new identifier field in the annotator dictionary that
is basically a unique hash produced by the annotator name, email,
affiliation, etc?

ejhumphrey:
[...]

PR #202, opened 2019-08-09
Support jsonschema>=3.0

### LINKED ISSUE #193: support jsonschema >= 3.0
Dropping this here until we have time to deal with it properly.  JAMS tests currently fail on jsonschema 3.0.0a2.

bmcfee:
This PR updates the jsonschema dependency to support >= 3.0, and fixes #193 .

The fix involves forcing the validator to use draft-04, which is now out of date.  But this stop-gap solution works for now.

We should have a separate issue, tied to #92 , which could involve updating the schema definition itself to draft 7.  There appear to be some good technical reasons for doing so (namely: relative uri references), but we can punt that down the road for now.

### DIFF
--- a/jams/core.py
+++ b/jams/core.py
@@ -581,7 +581,7 @@ class JObject(object):
         valid = True
 
         try:
-            jsonschema.validate(self.__json__, self.__schema__)
+            schema.VALIDATOR.validate(self.__json__, self.__schema__)
 
         except jsonschema.ValidationError as invalid:
             if strict:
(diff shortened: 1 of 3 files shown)


bmcfee:
Tagging @justinsalamon for a CR request -- nothing major here, everything passes and works as expected.

bmcfee:
For the record, this fix also seems to run just fine on jsonschema 2.6, so we don't lose any backward compatibility.  It does bump the dependency version, so that shouldn't be a problem, but it's good to know.

bmcfee:
tagging @beasteers for opinions

bmcfee:
> So I believe you said that Draft4 was the default before so that means we don't have to worry about falling back to previous drafts, correct?

Yup, that's correct.  The jams schema was written with draft4 (it even says so explicitly), so here we're just making sure that the right validator is being used.

PR #204, opened 2019-08-26
Boundary condition on trimming

### LINKED ISSUE #203: `jams.Annotation.trim()` doesn't include events with 0 duration that's on the trim boundary
This issue is about jams.Annotation.trim()

Reproduciable code is below:
```
import jams
print(jams.__version__)

ann = jams.Annotation(namespace='beat', duration=4)
ann.append(time=0, duration=0, value=1)
ann.append(time=1, duration=0, value=2)
ann.append(time=2, duration=0, value=3)

print('----------')
for obs in ann.trim(0, 2.1, strict=False):
    print(obs)
    
print('----------')
for obs in ann.trim(0, 2, strict=False):
    print(obs)
    
print('----------')
for obs in ann.trim(-1, 1.9, strict=False):
    print(obs)
```

Current Output:
```
0.3.4
----------
Observation(time=1.0, duration=0.0, value=2, confidence=None)
Observation(time=2.0, duration=0.0, value=3, confidence=None)
----------
Observation(time=1.0, duration=0.0, value=2, confidence=None)
----------
Observation(time=1.0, duration=0.0, value=2, confidence=None)
```

Desired Output:
```
0.3.4
----------
Observation(time=0.0, duration=0.0, value=1, confidence=None)
Observation(time=1.0, duration=0.0, value=2, confidence=None)
Observation(time=2.0, duration=0.0, value=3, confidence=None)
----------
Observation(time=0.0, duration=0.0, value=1, confidence=None)
Observation(time=1.0, duration=0.0, value=2, confidence=None)
Observation(time=2.0, duration=0.0, value=3, confidence=None)
----------
Observation(time=0.0, duration=0.0, value=1, confidence=None)
Observation(time=1.0, duration=0.0, value=2, confidence=None)
```

I took a look at the source code, and I think it's down to maybe adding some equal signs in this line: https://github.com/marl/jams/blob/6bc2d2e25f17860235dcd5c75fd348d3826f30b9/jams/core.py#L919
But I'm not sure if it will break other things.

Is this intended behavior? Right now it's impossible to slice or trim out the first beat of this annotation.

bmcfee:
This PR fixes #203 by including observations which have end time coincident with the trimming start time.

I'm still not 100% sure this is completely correct.  It works for @tomxi's example case, where there was a time=0, duration=0 observation being trimmed to start at 0.  This should clearly be included.

However, if we have an observation `[2, 2+1)` with a trim of `[3, 10)`, the modified logic will allow it to pass through, when it probably shouldn't, since `[2, 3) ∩ [3, 10) = {}`.  The ambiguity here comes from our special-case handling of duration-0 observations, which we treat as closed intervals to allow a (Nyquist-style) sampling interpretation.

Perhaps we should explicitly handle this case, by only allowing the boundary condition `end_time >= trim_start` if the observation duration is 0?

### DIFF
--- a/docs/jams_structure.rst
+++ b/docs/jams_structure.rst
@@ -29,7 +29,7 @@ Annotation
 
 Each annotation object contains the following properties:
     * ``namespace`` : a string describing the type of this annotation;
-    * ``data`` : a list of observations, each containing:
+    * ``data`` : a list of *observations*, each containing:
         * ``time`` : non-negative number denoting the time of the observation (in seconds)
         * ``duration`` : non-negative number denoting the duration of the observation (in seconds)
         * ``value`` : actual annotation (e.g., chord, segment label)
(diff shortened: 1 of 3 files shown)


justinsalamon:
Sounds like a reasonable solution to me (explicitly handling), given that for events with any non-zero duration this becomes a non-issue.

bmcfee:
Shall we merge this then?

tomxi:
Not sure if it counts for anything.... but I took a quick peek at the changed files and it all LGTM. Thanks for addressing this.

bmcfee:
Looks like I dropped the ball on this, but at ~3 months it looks fine to me.  :shipit:

PR #210, opened 2020-10-21
Allow to change the default indentation of the produced JSON.

hendriks73:
The default is 2, which can lead to fairly large bloat, when considering ten-thousands of JAMS (think global tempo estimates like https://github.com/tempoeval/tempo_eval/tree/main/annotations). Of course one can go the gzip/jamz route, which will create smaller files, but at the price of (easy/casual) editability.

Being able to simple pass indent=None is a compromise. It still allows manual editing, while also reducing size quite a bit.

### DIFF
--- a/jams/core.py
+++ b/jams/core.py
@@ -1741,7 +1741,7 @@ class JAMS(JObject):
 
         return self.annotations.search(**kwargs)
 
-    def save(self, path_or_file, strict=True, fmt='auto'):
+    def save(self, path_or_file, strict=True, fmt='auto', indent=2):
         """Serialize annotation as a JSON formatted stream to file.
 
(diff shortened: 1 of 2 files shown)


hendriks73:
Closed and re-opened PR to trigger Travis build.
Since Master has changed, it should run through this time.
The change is trivial.

justinsalamon:
Lgtm, @bmcfee any comments before I merge and bump the patch version?

bmcfee:
> Lgtm, @bmcfee any comments before I merge and bump the patch version?

LGTM, thanks!

Before bumping the version, I think it's worth combing through open issues and seeing if there's anything else that needs fixing in the short term.

PR #212, opened 2020-11-30
Build and test on Python 3.7 and 3.8

hendriks73:
Added python 3.7 and 3.8 build targets to `.travis.yml` and added these Python versions to `setup.py` classifiers.
This is to ensure JAMS can be used in these environments.


bmcfee:
LGTM, thanks!

hendriks73:
Thanks, Brian.

Issue #208, opened 2020-04-22
Next generation jams


bmcfee:
This issue is intended to consolidate many of the long-standing issues and offline discussions we've had around revising the jams specification for a variety of applications and use-cases.

---

## Goals of revising the schema

1. Migrate to a fully json-schema compliant spec #178 (instead of hybrid / dynamic namespaces).
    - This would facilitate a number of applications, like #19 (quick look / web view and edit), #40 (collection management), and #86 (fragmented storage).
2. Add versioning to the schema definitions.  This way, old files can still validate according to their specified jams version.  This in turn makes it easier to evolve the schema without breaking compatibility.
3. Simplify (and accelerate) the validation code from the python side.

### Revision phase 1: full json-schema definition

The first step is to move all namespace definitions into full jsonschema definitions.  In the proposed change, a namespace definition now becomes a secondary schema for the `Annotation` object.

`Annotation` objects must validate against both the template schema (our current annotation schema def), and *exactly one of* the pre-defined namespace schemas.  Each namespace schema defines an exact match on the `Annotation.namespace` field, in addition to whatever constraints placed on the `value` and `confidence` fields.

The `is_sparse` flag will be removed, as this is not part of jsonschema.  (We'll come back to this later).

This phase will complete #178 .

### Revision phase 2: hosted and versioned schema

Completing phase 1 will result in a fully json-schema compatible implementation of our specification, against which all current JAMS files should validate.

The next step (phase 2) is to place this schema under version control and host it remotely (e.g. `jams.github.io/schema/v0.3/schema.json`` or something).  We can then revise the schema to include a version number in its definition, so that jams files can self-identify which version they are valid under.

With the remote schema implementation, it should be possible/easy to promote all jams definitions to top-level objects, so that you can independently validate an `Annotation` or `FileMetadata` object without having it belong to a full JAMS file.

This phase will complete #86 and facilitate #40 , by allowing partial storage.

### Revision phase 3: extending the Annotation class

As mentioned in #24 , the current annotation structure might be a bit too rigid for more general media objects.  @justinsalamon and I discussed this offline, and arrived at the following proposal:

- Rename `Annotation` def to `IntervalAnnotation`, in which observations are `(time, duration, value, confidence)` tuples
- Add new annotation types 
  - `StaticAnnotation`: just `(value, confidence)`
  - `BoundingBoxAnnotation`: `(x, y, width, height, value, confidence)`
[...]

PR #213, opened 2021-01-02
JSONSchema compliance for namespaces

MCMcCallum:
- Addresses issue #208 phase 1
 - Updated JAMs schema to be json schenma compliant
 - Previously dense observation types are now entered as a single observation in the Annotation observation list, they no longer redefine the observation list itself, rather, just the elements within it. This means that all Annotations now have a list of observations, the elements of which can be dense observations themselves
 - the import_labs function has been moved from the util module to the core module. In a future PR it should be modified to be a constructor for JObjects
 - Fixed serialization JAMs for bools, lists and dicts
 - Updated jsonschema to latest version - necessary for determining namespaces based on a const namespace property
 - Local namespaces are no longer supported - a necessity to become jsonschema compliant
 - Version bumped
 - Tests updated to adhere to new schema for objects with DenseObservations
 - Removed spurious validation and __schema__ methods / attributes in jams.core module
 - add_namespace and is_dense is removed from the jams.schema module. All namespaces are now available in the jams.schema.NAMESPACES dictionary
 - schema for each observation type is now available in the jams.schema.OBSERVATIONS_SCHEMA dictionary


justinsalamon:
Hey! Is this in response to an issue? Was there a side convo I’m missing?

justinsalamon:
Ah, #208 

this seems like a pretty substantial design change 😬 we typically try to hash those out in issues before implementing 😅 

I believe @bmcfee has some fairly concrete thoughts on dense vs sparse annotation types.

My primary concern is ensuring major changes don’t  break Scaper.

Let’s discuss!

bmcfee:
> this seems like a pretty substantial design change grimacing we typically try to hash those out in issues before implementing sweat_smile

@MCMcCallum and I had actually hashed this out quite a bit before the implementation, but it was out-of-band.

Sorry that I haven't had time to dig into this PR thoroughly btw.  Teaching in a pandemic, you know :grimacing: 



> My primary concern is ensuring major changes don’t break Scaper.

The plan in #208 , phase 1, is to retain backward compatibility to the previous schema.  However, dynamic namespace support will eventually dropped, and scaper will likely need to adapt in response.

PR #215, opened 2023-11-01
Migrate deprecated imp package to importlib

### LINKED ISSUE #214: Install fails on Python 3.12: `imp` Module Missing
The `imp` package, used in `setup.py`, is no longer available in Python 3.12  (see [docu](https://docs.python.org/dev/whatsnew/3.12.html#imp)).

As a result, executing `pip install` under Python 3.12 fails with:

```
ModuleNotFoundError: No module named 'imp'
```

robaerd:
This PR fixes #214 by migrating from `imp` to `importlib`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,18 @@
 from setuptools import setup, find_packages
 
-import imp
+import importlib.util
+import importlib.machinery
 
-version = imp.load_source('jams.version', 'jams/version.py')
+
+def load_source(modname, filename):
+    loader = importlib.machinery.SourceFileLoader(modname, filename)
+    spec = importlib.util.spec_from_file_location(modname, filename, loader=loader)
+    module = importlib.util.module_from_spec(spec)
+    loader.exec_module(module)
+    return module
+
+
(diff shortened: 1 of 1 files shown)


bmcfee:
Thanks for this, and sorry for the slow response.  I'll merge this after we fix the CI to run on github actions so we can verify that it behaves correctly.

hyve9:
Just FYI, without this PR mirdata cannot be used with versions of python 3.11 or later

bmcfee:
Just noting that I'll try to get this squared away this week.  There are some blockers upstream that also needed to be updated (eg mir_eval) before this makes sense to do, but the work is mostly done at this point.

bmcfee:
Sorry, this got away from me :facepalm: 

The mir_eval reelase is still blocking CI updates, but this PR is okay for the time being.  Bigger picture, we can actually do away with this importlib stuff entirely by modernizing the packaging info to use setup.cfg and pull version from attributes, rather than doing import hackery.  But I can take that into a separate PR.

PR #217, opened 2024-07-04
Migrate CI to GitHub Actions

guillaumekh:
Hi,

I understand there was interest from the maintainer in migrating CI to GitHub Actions.

Having some familiarity with GitHub Actions, and needing to get this package otherwise  working with modern Python, I endeavored to give it a try.

It was more work than hoped for, and some more work is still needed to get this over the finish line.

Right now, a lot of tests are still failing but I lack the time to dig further, and am hamstrung by my near total lack of experience w/ Python.

Some notes:
- `numpy`@1.25.0 dropped support for Python <= 3.8
- `numpy`@1.20+ [broke](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated) `mir_eval`.
  - A fix was [merged](https://github.com/craffel/mir_eval/pull/353) in `mir_eval` _main_ branch but has not been released yet.
  - As a — really really bad — crutch, I've changed the dependency on `mir_eval` _main_ branch for the time being 🫠
  - As mentioned elsewhere here, a new `mir_eval` release is what we _really_ need. 
- `matplotlib`@3.9.0 [brakes](https://github.com/craffel/mir_eval/issues/383) `mir_eval`.
  - Did I mention we need a new `mir_eval` release? 
- i updated `pytest` hoping it would help solve failing tests but that's a HUGE v3—>v8 jump. I only made the bare minimum of fixes to accomodate this upgrade, and I would expect more work is needed.
- i wanted to add python 11/12 support but delayed this until we can already get tests runnning properly on 3.9/3.10

Would anyone be willing to try and pick things up where I left them?


bmcfee:
Thanks for all the work on this!  I'm a bit swamped over the next couple of weeks, but I'll try to pick this up.

bmcfee:
:wave: finally got mir_eval 0.8 out the door, so this should be unblocked!

giovana-morais:
i'm on it!

bmcfee:
> * for python 3.9, there is this weird namespace error. for a reason i don't quite understand, [`resources.files('jams.schemata.namespaces')`](https://github.com/stefansjs/jams/blob/9b53cee1fd657993bd3794d59df5dd3c37ee16f1/jams/__init__.py#L22), introduced in this [PR](https://github.com/marl/jams/pull/221/files#), returns None. the weird thing is that `resources.files` was introduced in python 3.9, so it should have the same behavior as the other versions.

That's ... weird.  I wonder if it has something to do with being two levels down?

> * when i try to rollback to the previous implementation using pkg_resources, the tests break for python 3.12 as `pkg_resources` becomes deprecated

Ahhh yeah, we hit this in librosa too.  Here's the way I patched it over there until we can bump the min python up to 3.9:

https://github.com/librosa/librosa/blob/3f88620f36d6726c3b656e8264828799ea7fa8fe/librosa/util/files.py#L38-L51

It's not exactly pretty, but it works.  A side benefit is explicitly separating the filename from the resource tree path, which is otherwise not exactly trivial because they both would split on the same token (`.`).  Eg:

https://github.com/librosa/librosa/blob/3f88620f36d6726c3b656e8264828799ea7fa8fe/librosa/util/files.py#L54-L57

If the problem you're hitting is due to ambiguous path splitting, this snippet might work for you.

OTOH, I'd be totally fine with dropping support for python < 3.9 in jams, and simplifying the resource loader accordingly.

bmcfee:
closing this one out in favor of the continuation #222 - @guillaumekh , don't worry, your commits are included there as well so we'll be able to track attribution.  Thanks again for your effort on getting this going!

PR #218, opened 2024-10-03
np.float_ -> np.float64

giovana-morais:
Hi all!
I tried to use mirdata with Python 3.10+ and jams was throwing an error regarding `np.float_`, so this PR changes that whenever it is necessary.

After doing the changes, I installed the library locally and had no problem with mirdata anymore. 
I was not able to run the tests though. It seems that [nose is unmaintained and not supported for Python 3.10+](https://github.com/SyneRBI/SIRF/issues/1148), so I'm not sure how to proceed here.


bmcfee:
Merging this since it's an obvious fix; will need to put some more work into modernizing CI to handle other stuff though

PR #219, opened 2024-12-02
setup.py fails on python > 3.4

stefansjs:
Python 3.4 deprecated the imp module. This PR replaces imp with an implementation using importlib. I also added a pyproject.toml so that pip can explicitly install setuptools when building the wheel.

### DIFF
--- a/jams/__init__.py
+++ b/jams/__init__.py
@@ -2,7 +2,8 @@
 """Top-level module for JAMS"""
 
 import os
-from pkg_resources import resource_filename
+from importlib import resources
+from itertools import chain
 
 # Import the necessary modules
 from .exceptions import *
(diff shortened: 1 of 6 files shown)


bmcfee:
Thanks for this.  I know we're long overdue for some updates - I swear it's on my radar as soon as the semester's over!

Meanwhile, would you mind minimizing this PR to just the issue noted here with importlib, and roll back the other changes?

stefansjs:
Yeah, sorry about that. I thought I was on another branch. I can't force-push to master, so I reverted the extra two commits

stefansjs:
Hmm, looks like this might actually be a duplicate (and identical implementation) to https://github.com/marl/jams/pull/215. Should I close this one?

FYI, I've got a bit of free time right now. If you're interested in delegating some of the PR management or bug fixing I've got the bandwidth.

bmcfee:
Thanks Stefan!  I'm happy to delegate some management right now.  One caveat though is that we're currently blocked on putting out a new release until mir_eval 0.8 goes out.  This is also on my stack, but we need to sort out some pypi credentialing and other management issues over there.  Software is complicated. :cold_sweat:

stefansjs:
Ok! I'm closing this PR because it's a duplicate. Let's sync up offline about what kinds of contributions I can do for these MIR repos.

PR #221, opened 2024-12-10
add support for Python3.12

stefansjs:
Adds support for python 3.12 by upgrading pytest to v8 and by pulling in changes that fix deprecated uses of pkg_resources.

This *should* unblock https://github.com/marl/jams/pull/217

### DIFF
--- a/jams/__init__.py
+++ b/jams/__init__.py
@@ -2,7 +2,8 @@
 """Top-level module for JAMS"""
 
 import os
-from pkg_resources import resource_filename
+from importlib import resources
+from itertools import chain
 
 # Import the necessary modules
 from .exceptions import *
@@ -18,14 +19,11 @@ from .schema import list_namespaces
 
 
 # Populate the namespace mapping
-for _ in util.find_with_extension(resource_filename(__name__, schema.NS_SCHEMA_DIR),
-                                  'json'):
-    schema.add_namespace(_)
+for ns in chain(*map(lambda p: p.rglob('*.json'), resources.files('jams.schemata.namespaces').iterdir())):
+    schema.add_namespace(ns)
 
 # Populate local namespaces
 
-try:
-    for _ in util.find_with_extension(os.environ['JAMS_SCHEMA_DIR'], 'json'):
-        schema.add_namespace(_)
-except KeyError:
-    pass
+if 'JAMS_SCHEMA_DIR' in os.environ:
+    for ns in util.find_with_extension(os.environ['JAMS_SCHEMA_DIR'], 'json'):
+        schema.add_namespace(ns)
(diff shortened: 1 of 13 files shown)


stefansjs:
This is a huge text change, but the short version is that xfail is both unsupported in decorators, and doesn't mean what we were using it to mean.

There is still a problem with `jams.sonify`, which I feel confident was not caused by these changes. I haven't attempted to fix them because they may imply a bigger problem with dependencies, and we should pin dependency versions correctly if that's the case. It seems as if some tests are expecting a different return type and test comparisons are failing. 

I hope somebody with more knowledge of the codebase than me can take a look and see if we need to pin a dependency version.

stefansjs:
Here's an example of the failures I'm still seeing https://github.com/stefansjs/jams/actions/runs/12251723660/job/34176899089#step:5:2797

```
=========================== short test summary info ============================
FAILED tests/test_sonify.py::test_duration[None-8000-chord] - TypeError: object of type 'numpy.float64' has no len()
FAILED tests/test_sonify.py::test_duration[None-11025-chord] - TypeError: object of type 'numpy.float64' has no len()
FAILED tests/test_sonify.py::test_duration[5.0-8000-chord] - TypeError: object of type 'numpy.float64' has no len()
FAILED tests/test_sonify.py::test_duration[5.0-11025-chord] - TypeError: object of type 'numpy.float64' has no len()
FAILED tests/test_sonify.py::test_duration[1.0-8000-chord] - TypeError: object of type 'numpy.float64' has no len()
FAILED tests/test_sonify.py::test_duration[1.0-11025-chord] - TypeError: object of type 'numpy.float64' has no len()
FAILED tests/test_sonify.py::test_note_hz - TypeError: object of type 'numpy.float64' has no len()
FAILED tests/test_sonify.py::test_note_hz_nolength - TypeError: object of type 'numpy.float64' has no len()
FAILED tests/test_sonify.py::test_note_midi - TypeError: object of type 'numpy.float64' has no len()
FAILED tests/test_sonify.py::test_chord[C:maj/5-2.0-8000-chord] - TypeError: object of type 'numpy.float64' has no len()
FAILED tests/test_sonify.py::test_chord[C:maj/5-2.0-8000-chord_harte] - TypeError: object of type 'numpy.float64' has no len()
================ 11 failed, 686 passed, 1815 warnings in 8.44s =================
```

bmcfee:
Thanks so much @stefansjs !

I've run this locally and can confirm the sonification failures.  These need to be fixed upstream, so not much we can do here for now.  I'm happy to merge this now so we can get CI back on its feet and then make some progress.

bmcfee (approved):

